PR #2, opened 2021-11-30
Fix Deck OAuth

timebertt:
Switch from App auth to robot PAT in deck (used for resolving teams for the rerun auth configs).
Currently, OAuth flow isn't implemented in combination with app auth: https://github.com/kubernetes/test-infra/issues/23870

With this, deck OAuth flow is fixed, i.e. one can now use the personal PR dashboard (https://prow.gardener.cloud/pr) and the rerun button triggers new jobs (if you're part of the required team).


rfranzke (commented):
/lgtm

PR #10, opened 2021-12-03
Add gardener e2e test (postsubmit)

timebertt:
/kind enhancement
Adds a job (postsubmit for now), that runs the new e2e test with provider-local, ref https://github.com/gardener/gardener/issues/5024


rfranzke (commented):
nice!

/lgtm

PR #40, opened 2022-01-18
Deploy ingress-nginx controller

### LINKED ISSUE #37: Replace Gardener ingress-nginx addon by a more recent version of ingress-nginx
**What would you like to be added**: Install a self-managed (and more recent) version of ingress-nginx in the prow cluster

**Why is this needed**: Using Gardener ingress-nginx addon is not recommended

oliver-goetz:
This PR includes the most recent version of nginx-ingress controller in a configuration with 2 replica. Its pods are assigned to worker nodes control-plane type only. It is created from [nginx-ingress helm chart](https://github.com/kubernetes/ingress-nginx/tree/main/charts/ingress-nginx) in v4.0.16
There is a generator script at `/config/scripts/generate-ingress-nginx-deployments.sh` which is able to update the deployment spec from the latest version of the chart using the same configuration.

ingress-nginx addons on trusted and workload clusters have been deactivated.
There is no need for an ingress on the `prow-work` cluster, hence it was deactivated without a replacement.

Fixes #37

`deployer` SA needed additional permission for creating jobs, webhook and poddisruptionbudgets. 

**Release note**:

```feature user
NONE
```


config/scripts/generate-ingress-nginx-deployments.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke, line 2: Nit: 2022

rfranzke (commented):
Skimmed through it - looks great!
/assign @timebertt for final feedback and approval

config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
```diff
@@ -0,0 +1,856 @@
...
+              containerPort: 8443
+              protocol: TCP
+          volumeMounts:
+            - name: webhook-cert
+              mountPath: /usr/local/certificates/
+              readOnly: true
+          resources: 
+            requests:
+              cpu: 100m
+              memory: 90Mi
```
timebertt, line 472: Can we give this one a bit more memory?
```
$ k -n ingress-nginx top po
NAME                                           CPU(cores)   MEMORY(bytes)
ingress-nginx-controller-75d9bf7d6b-rlp9c      3m           109Mi
ingress-nginx-controller-75d9bf7d6b-z58wp      3m           110Mi
ingress-nginx-defaultbackend-8d9b797b7-2xbz2   1m           4Mi
```
And maybe add limits as well.

config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
```diff
@@ -0,0 +1,856 @@
...
+  replicas: 1
+  revisionHistoryLimit: 10
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: ingress-nginx
+        app.kubernetes.io/instance: ingress-nginx
+        app.kubernetes.io/component: default-backend
+    spec:
+      containers:
```
timebertt, line 520: Requests and limits for this would be nice.

config/scripts/generate-ingress-nginx-deployments.sh
```diff
@@ -0,0 +1,76 @@
...
+cat <<EOF > $SCRIPT_DIR/../prow/cluster/ingress-nginx/00_ingress-nginx-namespace.yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ingress-nginx
+EOF
+
+echo "Templating ingress-nginx"
+# Check values.yaml for help https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml
+helm template -n ingress-nginx ingress-nginx ingress-nginx/ingress-nginx -f - > $SCRIPT_DIR/../prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml <<EOF
```
timebertt, line 42: Nice. I like the pragmatic approach here.
Can we move the script to `config/prow/cluster/ingress-nginx`, though? (just like `config/prow/cluster/monitoring/prometheus-operator-crds/update-crds.sh`)
Also, can you move out the inline values to a `values.yaml` file.
This makes things easier to find, understand and look at :)

config/scripts/generate-ingress-nginx-deployments.sh
```diff
@@ -0,0 +1,76 @@
...
+controller:
+  replicaCount: 2
```
timebertt, line 44: Let's go for 3 replicas just like for `hook` and `deck`

timebertt (changes requested):
This looks great, thank you!
A few minor suggestions inline :)

config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
oliver-goetz: I increased the memory request of ingress-controller.
However, the authors of the nginx helm chart recommend [not setting a resource limit](https://github.com/kubernetes/ingress-nginx/blob/d16e0dec4865cbd5b937dbffb957423c308331c5/charts/ingress-nginx/values.yaml#L343-L346) and I don't know better than they do

timebertt: Hmm, what I read from the referenced blog post:

> If you are using Kubernetes, use whole CPU quotas, as this decreases the number of schedulable CPUs available to the cgroup

So what do you think about the following limits: `cpu: 1, memory: 500Mi`?

Setting limits generally makes sense to avoid pods that are too resource hungry or leaking memory to bring down other workload or even system components on the same node. 
This is a common issue that we have observed quite often in the past. Hence, I would still set limits for these components. :)

timebertt (commented):
Changes look good. Left one comment about the controller's resource limits

config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
oliver-goetz: ok, let's go for it 😄

timebertt (approved):
Nice, let's go!
/lgtm

PR #45, opened 2022-01-19
Simplify prow trusted cluster setup

### LINKED ISSUE #44: Reduce complexity of cluster setup
Using a dedicated node type for prow jobs in the trusted cluster looks like over-engineering.
There is a dedicated cluster for untrusted jobs which are the most compute and memory demanding tasks too.

Thus, we should apply those changes:
- Using only a single type of node for trusted prow cluster with a minimum autoscaling limit of 2
- Distributing apps with replica > 1 across all nodes to improve resilience when nodes fail

**Why is this needed**:
A simplified setup is less error-prone and would save money because we can use smaller compute units for trusted prow cluster.

oliver-goetz:
- Remove node type controlplane from prow-trusted cluster
- Resize worker nodes to `e2-standard-4`
- Increase the minimum number of worker nodes to 2
- Switch to containerd
- Add PodAntiAffinity rules of type `preferredDuringSchedulingIgnoredDuringExecution` to Pods with replica > 1 to spread them across multiple nodes

Fixes #44 

Controlplane worker nodes are removed manually when the PR is merged. Otherwise, ingress could not be scheduled anymore in the meantime.

**Release note**:

```feature user
NONE
```

### DIFF
--- a/config/clusters/shoot-workload.yaml
+++ b/config/clusters/shoot-workload.yaml
@@ -56,6 +56,8 @@ spec:
       maxUnavailable: 0
       systemComponents:
         allow: true
+      cri:
+        name: containerd
       volume:
         size: 100Gi
         type: pd-ssd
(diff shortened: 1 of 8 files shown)


timebertt (commented):
Nice PR, thank you!
/lgtm
/approve

PR #56, opened 2022-01-26
Add gcsweb deployment

### LINKED ISSUE #55: Improve visibility of jobs artifacts visible with gcsweb
/kind enhancement
/assign oliver-goetz

Deployment of [gcsweb](https://github.com/kubernetes/k8s.io/tree/main/apps/gcsweb)

**Why is this needed**:
Linking jobs with job artifacts which are already pushed to gcs buckets on prow.gardener.cloud.

oliver-goetz:
/kind enhancement

Deploy gcsweb in a HA setup with 3 replicas and enable it in prow configuration.

Fixes #55 


config/prow/cluster/gcsweb_ingress.yaml
```diff
@@ -0,0 +1,28 @@
...
+metadata:
+  name: gcsweb
+  namespace: gcsweb
+  labels:
+    app: gcsweb
+  annotations:
+    cert.gardener.cloud/issuer: ci-issuer
+    cert.gardener.cloud/purpose: managed
+    dns.gardener.cloud/class: garden
+    dns.gardener.cloud/dnsnames: gcsweb.gardener.cloud
```
rfranzke, line 12: Should we rather call this `gcsweb.prow.gardener.cloud`? (same below)

config/prow/config.yaml
```diff
@@ -45,7 +45,7 @@ sinker:
 deck:
   spyglass:
     size_limit: 100000000 # 100MB
-#    gcs_browser_prefix: https://gcsweb.k8s.io/gcs/
+    gcs_browser_prefix: https://gcsweb.gardener.cloud/gcs/
```
rfranzke, line 48: same here?

config/prow/cluster/gcsweb_ingress.yaml
timebertt: Actually, this is not prow-specific and rather a general GCS bucket browser. We could potentially use it for other things as well later on.
We were thinking about `gcsweb.prow.gardener.cloud` as well, but decided for `gcsweb.gardener.cloud` in the end, because of this and symmetry with `gcsweb.k8s.io`.
WDYT?

rfranzke: OK, understood. Then let's keep it as suggest. Thanks for explaining!

config/prow/cluster/gcsweb_deployment.yaml
```diff
@@ -0,0 +1,64 @@
...
+          image: k8s.gcr.io/gcsweb:v1.1.0
+          args:
+            - -upgrade-proxied-http-to-https
+            # buckets owned by gardener
+            - -b=gardener-prow
+            - -p=8080
+          ports:
+            - containerPort: 8080
+              protocol: TCP
+          resources:
```
timebertt, line 35: Can you add some minimal requests?

timebertt (commented):
Thank you, just one nit from my site :)

config/prow/cluster/gcsweb_deployment.yaml
oliver-goetz: ok :)

timebertt (commented):
Thank you!
/lgtm
/approve

PR #93, opened 2022-02-17
Move periodic check for cla status from prow job into cla-assistant plugin

oliver-goetz:
…

/kind enhancement

It turned out that it is not efficient to use a commenter prow job for periodic checks of CLA status.
We cannot influence the availability of cla-assistant.io. Because of that it is difficult to set a reasonable frequency for the checks.
A high frequency could result in a "/cla comment spam" by the commenter prow job, in case cla-assistant.io is not responsive for a longer time.
With a low frequency on the other hand it would take a long time until the recheck is started.

cla-assistant plugin includes a periodic check anyhow. Thus, the effort of adding the recheck function here is pretty low.

Fixes #


timebertt (approved):
Thanks!
/lgtm

PR #105, opened 2022-02-25
Build gardener images using prow, tekton and kaniko

oliver-goetz:
/kind poc

In this approach gardener container images are built using a combination of prow, tekton and kaniko.
The build pipeline uses caching and parallel builds in order to improve the speed of image builds.

These components are deployed:
- tekton pipeline - [install](https://github.com/tektoncd/pipeline/blob/main/docs/install.md)
- [tekton dashboard](https://github.com/tektoncd/dashboard) in read only mode
- [prow pipeline](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/pipeline)
- apps from tekton catalog to for [cloning git repositories](https://github.com/tektoncd/catalog/blob/main/task/git-clone/0.5/git-clone.yaml) and for [building images with kaniko](https://github.com/tektoncd/catalog/blob/main/task/kaniko/0.5/kaniko.yaml)

Those components are used in the [tekton build pipeline for gardener](https://github.com/oliver-goetz/ci-infra/blob/tekton/config/prow/cluster/tekton/kaniko-pipelines/pipeline-gardener-gardener.yaml).
The pipeline is started by a [prow job](https://github.com/oliver-goetz/ci-infra/blob/tekton/config/jobs/gardener/build-gardener-images.yaml).
There is the [tekton dashboard](https://tekton.prow.gardener.cloud/#/about) running for observability reasons because the capabilities of prow tekton plugin is very limited

**Findings**
- Build performance (using kaniko) depends heavily on the cpu type of the build cluster. Using an n2d-standard-16 machine instead of a e2-standard-16 reduces the build time by at least 30%
- Build steps which use cached layers are limited by disk IO.
  - Parallel builds of targets using cached layers are faster when using SSD drives on node and PVC only.
  - An over-provisioning of the SSD of the builder node decreases build times 

**Current limitations**
- prow plugin for integration of tekton pipeline seems to be in "maintenance mode" and lacks of observability functions
  - No logs are displayed in deck
  - Not enabled for prow sidecar, so no build artefacts are uploaded to gcloud storage bucket
 - pods of completed kaniko builds runs are not deleted automatically by tekton
   - this leads to increasing consumption of disk space on the builder node
   - cleaning up kaniko pods might be easy, but tekton seems to get the logs from those pods. Thus, deleting the pods means deleting the build logs 
 - kaniko build cache in container registry is not cleaned automatically   

This is a draft PR to evaluate different build options for gardener repositories.
A second proposal using a custom app to run the build steps in right sequence will be created soon.

### DIFF
--- a/config/clusters/shoot.yaml
+++ b/config/clusters/shoot.yaml
@@ -13,24 +13,26 @@ spec:
   cloudProfileName: gcp
   dns:
     providers:
-    - type: aws-route53
-      secretName: aws-route53-ci
+      - secretName: aws-route53-ci
+        type: aws-route53
   extensions:
-  - type: shoot-cert-service
-    providerConfig:
-      apiVersion: service.cert.extensions.gardener.cloud/v1alpha1
-      kind: CertConfig
-      issuers:
-      - name: ci-issuer
-        server: 'https://acme-v02.api.letsencrypt.org/directory'
-        email: tim.ebert@sap.com
+    - type: shoot-cert-service
+      providerConfig:
+        apiVersion: service.cert.extensions.gardener.cloud/v1alpha1
+        issuers:
+          - email: tim.ebert@sap.com
+            name: ci-issuer
+            server: 'https://acme-v02.api.letsencrypt.org/directory'
+        kind: CertConfig
+  hibernation:
+    enabled: false
   kubernetes:
     allowPrivilegedContainers: true
     kubeAPIServer:
       enableBasicAuthentication: false
     kubeProxy:
       mode: IPTables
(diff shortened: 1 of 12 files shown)


timebertt:
Thanks again @oliver-goetz for experimenting with this approach and providing the rough implementation.
Should we `/close` this PR now, as we have decided on rather pursuing https://github.com/gardener/ci-infra/pull/116?

oliver-goetz:
yes, closing this is a good idea.
I already removed the evaluation deployments we did for this PR from our cluster.
/close

PR #117, opened 2022-03-10
Prow housekeeping with small enhancements

oliver-goetz:
/kind enhancement

Includes several small enhancements we collected over the time
- creation of `deployer` service account and roles & permissions is moved from `deploy.sh` to `bootstrap-config.sh`
  - `deploy.sh` is started by prow job `post-ci-infra-deploy-prow` using service account `deployer`
  - as a service account can only limit its permissions and never extend them, it makes more sense to run this in a script which is invoked manually
- auto-deployment of prow autobump contents is now started in the morning. The majority of PRs in kubernetes/test-infra are merged are merged during US working hours, so if we catch up with reality in the morning we should be good for the day.
- adding additional trigger files for running go tests of ci-infra components where changes could make tests fail
- prow clusters have been updated over the time. The configuration files in github are now catching up with reality

Fixes #


config/jobs/ci-infra/ci-infra-presubmits.yaml
```diff
@@ -27,7 +27,7 @@ presubmits:
         - --warnings=duplicate-job-refs
   - name: pull-ci-infra-prow-go-tests
     cluster: gardener-prow-build
-    run_if_changed: '^(prow\/*)'
+    run_if_changed: '^(prow\/*|Dockerfile|Makefile|go.(mod|sum))'
```
timebertt, line 30: `vendor` is missing.
Maybe it doesn't bring much value to restrict the test to specific paths. The overhead of always running it shouldn't be too high.

config/prow/deploy.sh
```diff
@@ -124,14 +124,14 @@ kubectl config use-context gardener-prow-trusted
 kubectl apply --server-side=true -f "$SCRIPT_DIR/cluster/prowjob_customresourcedefinition.yaml"
 
 for c in "${prow_components[@]}"; do
-  kubectl apply --server-side=true -f "$SCRIPT_DIR/cluster/$c"
+  kubectl apply -f "$SCRIPT_DIR/cluster/$c"
```
timebertt, line 127: Why?

config/jobs/ci-infra/ci-infra-presubmits.yaml
oliver-goetz: I agree and removed the condition

config/prow/deploy.sh
oliver-goetz: When I changed it and ran the script manually prow self deployment [started to fail](https://prow.gardener.cloud/?job=post-ci-infra-deploy-prow). Apparently `deployer` service account does not have all permissions it needs to deploy prow completely. 
Afaik server-side deployment removes the last-applied-version annotations. Thus, deployer has to change every object when switching the server-side parameter and the lack of permissions becomes visible.
As it is related to roles and permissions I prefer to investigate changes in a separate pull request.

timebertt: Alright. Then let's do the switch to SSA later on.

timebertt (commented):
Thanks!
/lgtm
/approve

PR #122, opened 2022-03-14
Improve scheduling of test jobs

oliver-goetz:
/kind enhancement

This PR includes two improvements of scheduling of the test jobs we introduced recently
- introduce periodic test jobs running every 4 hours
- optimize resource request of test jobs

*Resource request optimization:*
In current configuration the 3 test of a PR request more CPU than an entire node provides. Thus, in many cases the cluster has scale up, before the pods of all tests could be started.
Our current worker machine provides 16 CPUs of which 15.9 are allocatable. About 1.1 CPUs are allocated for system components and 14.8 remain for the test pods.
```
Capacity:
  cpu:                16
  ephemeral-storage:  102033388Ki
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             65867564Ki
  pods:               110
Allocatable:
  cpu:                15920m
  ephemeral-storage:  99258079769
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             64716588Ki
  pods:               110
```
We would like to optimize the processing time of the combination of our 3 PR tests while trying to avoid the cluster to scale up and down for a short time.
The proposed CPU request looks like this:
- `pull-gardener-e2e-kind`: 5 CPUs
  - this job consumes only little CPU resources most of the time, but has spikes where it his its current CPU limit of 7
  - the test normally runs 20-23min which is the test with the longest runtime we have. In order to extend runtime by CPU throttling it is getting a relatively high CPU request
- `pull-gardener-unit`: 5 CPUs
  - this job could consume the entire CPU of the node for most of the time
  - it is running for 9-10 minutes. A slightly increased runtime because of CPU throttling might be more efficient than waiting for a new node to come up.
- `pull-gardener-integration`: 4 CPUs
  - this job could consume the entire CPU of the node for most of the time
  - It is usually running for about 2 minutes, so the effect of CPU throttling would be even smaller 

Sidecars use the default CPU resource request of 0.1. Thus, our 3 test jobs will request a total amount of 14.3 CPUs.
There are 0.5 CPUs left which could be used for the small prow jobs in worker cluster.

All jobs will have no CPU limit to be able to consume all the CPUs of the worker nodes.
Memory limits are enforced. They are not very strict as our worker nodes usually have more than enough memory.

*Timeouts of E2E test are decreased to 60 minutes, which is about 3x their normal runtime.*

Fixes #

### DIFF
--- a/config/jobs/ci-infra/ci-infra-periodics.yaml
+++ b/config/jobs/ci-infra/ci-infra-periodics.yaml
@@ -258,3 +258,31 @@ periodics:
       - --warnings=validate-urls
       - --warnings=unknown-fields
       - --warnings=duplicate-job-refs
+- name: ci-ci-infra-prow-go-tests
+  cluster: gardener-prow-build
+  interval: 4h
+  extra_refs:
+  - org: gardener
+    repo: ci-infra
+    base_ref: master
+  reporter_config:
+    slack:
+      channel: "gardener-prow-alerts"
+  decorate: true
+  annotations:
+    description: Runs go tests for prow developments in ci-infra 
+  spec:
+    containers:
+    - image: golang:1.17.7
+      command:
+      - make
+      args:
+      - check
+      - test
+      - verify-vendor
+      resources:
+        limits:
+          memory: 4000Mi
+        requests:
+          cpu: 4
+          memory: 2000Mi
--- a/config/jobs/ci-infra/ci-infra-periodics.yaml
+++ b/config/jobs/ci-infra/ci-infra-periodics.yaml
@@ -282,7 +282,7 @@ periodics:
       - verify-vendor
       resources:
         limits:
-          memory: 4000Mi
+          memory: 4Gi
         requests:
           cpu: 4
-          memory: 2000Mi
+          memory: 2Gi
--- a/config/jobs/gardener/gardener-e2e-kind.yaml
+++ b/config/jobs/gardener/gardener-e2e-kind.yaml
@@ -45,10 +45,10 @@ presubmits:
           privileged: true
         resources:
           limits:
-            memory: 18000Mi
+            memory: 18Gi
           requests:
             cpu: 5
-            memory: 9000Mi
+            memory: 9Gi
         env:
         - name: SKAFFOLD_UPDATE_CHECK
           value: "false"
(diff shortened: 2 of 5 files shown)


oliver-goetz:
Pre submit jobs seem to run automatically only if they one of those items: `always_run`, `run_if_changed`, `skip_if_only_changed`

config/jobs/gardener/gardener-e2e-kind.yaml
```diff
@@ -45,10 +45,9 @@ presubmits:
           privileged: true
         resources:
           limits:
-            cpu: 7
-            memory: 9000Mi
+            memory: 18000Mi
           requests:
-            cpu: 7
+            cpu: 5
             memory: 9000Mi
```
timebertt, line 51: Go with `Gi` here?

config/jobs/gardener/gardener-pull-tests.yaml
```diff
@@ -45,6 +53,69 @@ presubmits:
         args:
         - test-integration
         resources:
+          limits:
+            memory: 4000Mi
           requests:
-            cpu: 6
+            cpu: 4
             memory: 2000Mi
+periodics:
```
timebertt, line 61: Maybe the file should be renamed to `gardener-tests.yaml`? Now it doesn't contain just PR jobs anymore.
Otherwise, you could split it up into `gardener-unit.yaml` and `gardener-integration.yaml` and include the corresponding presubmit and periodic.

timebertt (commented):
Thank you!
Generally lgtm, just a few nits

config/jobs/ci-infra/ci-infra-periodics.yaml
```diff
@@ -258,3 +258,31 @@ periodics:
...
+- name: ci-ci-infra-prow-go-tests
+  cluster: gardener-prow-build
+  interval: 4h
+  extra_refs:
+  - org: gardener
+    repo: ci-infra
+    base_ref: master
+  reporter_config:
+    slack:
+      channel: "gardener-prow-alerts"
```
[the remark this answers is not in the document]
oliver-goetz: It is on purpose 😄 
The idea behind it is using `gardener-test-failures` for errors from gardener components only.
This test is an prow internal development That's why I used the `gardener-prow-alerts' channel here

timebertt (commented):
Awesome, thanks!
/lgtm
/approve

PR #148, opened 2022-03-30
Only include desired branches in protection

timebertt:
/kind bug

`branch-protector` is iterating over branches in a repository and creating branch protections for each one.
We should only include well-known branches, to not protect branches from upgrade PRs like https://github.com/gardener/gardener/pull/5693. Otherwise, these cannot be deleted anymore.

Fixes #


rfranzke (commented):
/lgtm

PR #149, opened 2022-03-30
Run release tests on release branches only

oliver-goetz:
/kind bug

Release tests should run on release branches only.
In previous configuration, they ran on branches created by `gardener-robot-ci-x` too


timebertt (approved):
Thanks!
/lgtm

PR #153, opened 2022-03-30
gardener branding

oliver-goetz:
/kind enhancement

Add gardener icon and colors to https://prow.gardener.cloud 😄


config/prow/config.yaml
```diff
@@ -91,6 +91,10 @@ deck:
     '*':
       github_team_ids:
       - 5415312 # ci-infra-maintainers
+  branding:
+    header_color: "#0b8062"
+    logo: "/static-files/gardener.svg"
+    favicon: "/static-files/gardener-favicon.png"
```
timebertt, line 97: From https://github.com/kubernetes/test-infra/blob/1a17f7a6406878ed2f53a9b6865bf227c7d8c394/prow/cmd/deck/main.go#L1588-L1597 it looks like deck is trying to serve the `/favicon.ico` path from its own filesystem.
Maybe we have to mount it into deck as well for this to work.

timebertt (commented):
That's awesome. I like it! 👍

config/prow/config.yaml
oliver-goetz: Probably yes. Why is it handled differently than the logo? 😓 
ok, then let us mount everything into deck. My initial idea of using an own webserver was, that it does not fail silently in case the directory structure of deck changes.

timebertt (approved):
Lovely!
/lgtm

PR #154, opened 2022-03-30
Update nginx to v1.1.2

oliver-goetz:
/kind enhancement

Update ingress-nginx from v1.1.1 to v1.1.2

### DIFF
--- a/config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
+++ b/config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
@@ -4,10 +4,10 @@ apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   labels:
-    helm.sh/chart: ingress-nginx-4.0.16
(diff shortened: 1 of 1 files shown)


config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
```diff
@@ -4,10 +4,10 @@ apiVersion: policy/v1
 kind: PodDisruptionBudget
```
timebertt, line 4: This is generated from `config/prow/cluster/ingress-nginx/helm/generate-ingress-nginx-deployments.sh`, correct?

I would suggest to pin the chart version in the script, so that it's reproducible.

oliver-goetz: Yes, it is.
My idea having a script and a fixed yaml description and a script was using the yaml for a reproducible deployment and using the script to update this yaml to the most recent version of the chart.

timebertt (commented):
/lgtm
/approve

PR #155, opened 2022-03-31
Fix settings for concourse (3)

timebertt:
/kind bug

- don't send Slack warnings when concourse pushes releases and upgrade PRs
- don't enforce admins to push to protected branches (seems like this is the only way to allow concourse to push release commits – GitHub branch protection is super confusing, I still don't understand why)

Follow-up on #151, which has not helped unfortunately


oliver-goetz:
A clean way would be changing the behaviour of `gardener-robot-ci-xyz` to create PRs we could auto-merge instead of direct commits to the release branches. However, this would be a mid-term solution requiring changes in concourse too.

/lgtm

PR #167, opened 2022-04-06
Disable parallel builds / don't build ci-infra images on config changes

oliver-goetz:
/kind bug

Two small fixes regarding builds and autodeployment
- Do not build ci-infra images on changes in entire `config/`
  - Currently autobumping jobs triggers a build which triggers a new autobump which triggers a new build....
- Disable parallel builds for the same repository
  - parallel builds could mess up fixed tags like `latest` in case two commits are pushed at about the same time. With parallel builds it is not deterministic which build is tagged as `latest`

### DIFF
--- a/config/jobs/ci-infra/build-golang-test-image.yaml
+++ b/config/jobs/ci-infra/build-golang-test-image.yaml
@@ -8,6 +8,7 @@ postsubmits:
     annotations:
       description: Build golang-test image on master branch
     decorate: true
+    max_concurrency: 1
     reporter_config:
       slack:
         channel: "gardener-prow-alerts"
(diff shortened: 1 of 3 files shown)


config/jobs/ci-infra/build-golang-test-image.yaml
```diff
@@ -8,6 +8,7 @@ postsubmits:
     annotations:
```
timebertt, line 8: The problem here is: we have to trigger this job manually, if we only change the build arg value in the job config but nothing else. 
We might need to consider putting tags and args into a file under `/images` or something similar to what test-infra does: https://github.com/kubernetes/test-infra/tree/master/images/bazel
But we can run with this for now, I guess.

timebertt (approved):
/lgtm

PR #181, opened 2022-04-08
Build gardener/gardener dev images

### LINKED ISSUE #180: Build gardener/gardener images with prow/kaniko
/kind enhancement

Now that the prow image-builder app is completed and is building ci-infra apps productively and gardener apps in a test environment, we could think about the next steps to build gardener/gardener images in prow.

We could proceed in two steps:
1. Use prow for test builds replacing `concourse-ci/publish` job
2. Include prow image build into release pipeline

*Step 1: test builds*

When investigating the I noticed that the images built by `concourse-ci/publish` PR  test are pushed to `eu.gcr.io/gardener-project/gardener` which is the same location where we are pushing our release images too.
This does not look ideal from my point of view, because
- we cannot guarantee that those image come from a trusted origin
- I question, that we really need to push those images anywhere. From my perspective these are primary tests if the PR could be built at all.

I would like to propose a slightly different setup for prow
- deactivate `concourse-ci/publish` PR test
- create a new presubmit job for gardener/gardener which runs a plain kaniko pod without any target but with the [--no-push](https://github.com/GoogleContainerTools/kaniko#--no-push) flag
  - the pod will build the entire Dockerfile
  - it neither uses cache nor pushes the resulting images anywhere, so it can run in safely the prow-work cluster without any credentials for gcr.io
  - the goal of this job is to find out, if we are able to build the PR
- [Change the destination](https://github.com/gardener/ci-infra/blob/master/config/jobs/gardener/build-gardener-images.yaml#L24-L25) of `post-gardener-build-images` job to `eu.gcr.io/gardener-project/gardener`
  - we will build the current state of master branch
  - exclude commits which are changing the `VERSION` file only, that we do not build release images

Those are the gains from my point of view:
- we do not push images from "untrusted" origins to `eu.gcr.io/gardener-project/gardener` anymore
- we are able to validate if the prow images are working more easily
- we already reduce the load of our concourse systems

*Step 2: release builds*

After we verified that our builds are working as expected we can start building our release images in prow.

- create a build prow job similar to `post-gardener-build-images` for the release branches which is triggered by changes to `VERSIONS` file
- deactivate build in concourse pipeline but keep the rest
- `gardener-robot-ci-*` would initiate the release build by pushing a new `VERSION` file to the release branches

*Step 2 is still a draft. I cannot see the entire concourse release pipeline yet.*

**Why is this needed**:
Reduce load on concourse pipelines.
Improve development experience.
Build images of a trusted origin only.

oliver-goetz:
/kind enhancement

Builds gardener/gardener dev images in prow.

presubmit jobs:
- `pull-gardener-verify-image-build` builds all targets of Dockerfile but is not pushing the results anywhere
- `pull-gardener-publish-test-images` is an optional job to build and publish images of an PR to `eu.gcr.io/gardener-project/gardener`
  - it it started on request in a PR `/test pull-gardener-publish-test-images` and takes about 5 minutes
- Both jobs exist for release branches too `pull-gardener-verify-image-build-release`, `pull-gardener-publish-test-images-release`

postsubmit jobs:
- `post-gardener-build-dev-images` builds and publishes images from master branch to `eu.gcr.io/gardener-project/gardener`
  - is not running when only VERSION file is changed to not interfere release builds

This approach is slightly different than the current setup, but still allows to build test images from a PR.
Another option could be changing `pull-gardener-publish-test-images` to a non-optional prow job, to publish images from each commit of any PR and remove `pull-gardener-verify-image-build` 

A minor change is updating kaniko to the latest patch level 1.8.1.

Fixes #180 partly

`concourse-ci/publish` should be deactivated in gardener/gardener when merging this PR

### DIFF
--- a/config/jobs/gardener/build-gardener-dev-images.yaml
+++ b/config/jobs/gardener/build-gardener-dev-images.yaml
@@ -1,0 +1,65 @@
+postsubmits:
+  gardener/gardener:
+  - name: post-gardener-build-dev-images
+    cluster: gardener-prow-trusted
+    skip_if_only_changed: '^VERSION'
+    branches:
+    - ^master$
+    annotations:
+      description: Gardener development image build on master branch
+    decorate: true
+    max_concurrency: 1
+    spec:
+      serviceAccountName: image-builder
+      containers:
+      - name: image-builder
+        image: eu.gcr.io/gardener-project/ci-infra/image-builder:v20220407-1704ddf
+        command:
+        - /image-builder
+        args:
+        - --log-level=info
+        - --docker-config-secret=gardener-prow-gcr-docker-config
+        - --registry=eu.gcr.io/gardener-project/gardener
+        - --cache-registry=eu.gcr.io/gardener-project/ci-infra/kaniko-cache
+        - --target=apiserver
+        - --target=controller-manager
+        - --target=scheduler
+        - --target=gardenlet
+        - --target=admission-controller
+        - --target=seed-admission-controller
+        - --target=resource-manager
+        - --target=gardener-extension-provider-local
+        - --add-version-tag=true
+        - --add-version-sha-tag=true
+        # image-builder is the pod which is "scheduled" to a node. The pods created by image-builder have an affinity rule
+        # which schedules them to the same node as their parent image-builder. This needs to be done, that PVCs could be mounted
+        # to multiple build pods in parallel.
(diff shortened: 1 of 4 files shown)


timebertt (commented):
Note: we need make `councourse-ci/publish` not required anymore in branch protection.

timebertt (commented):
/lgtm
/hold
for @oliver-goetz

PR #184, opened 2022-04-11
Increase memory limits for gardener integration tests

timebertt:

Increase memory limits for gardener integration tests

Integration test jobs got `OOMKilled`:
https://prow.gardener.cloud/view/gs/gardener-prow/pr-logs/pull/gardener_gardener/5784/pull-gardener-integration/1513447190545043456
https://prow.gardener.cloud/view/gs/gardener-prow/pr-logs/pull/gardener_gardener/5784/pull-gardener-integration/1513454017219399680

Seems, that executing tests with `ginkgo` instead of `go test` temporarily consumes more memory while compiling (`ginkgo` probably just doesn't have as sophisticated resource management as `go tests`).
ref https://github.com/gardener/gardener/pull/5784

### DIFF
--- a/config/jobs/gardener/gardener-integration-tests.yaml
+++ b/config/jobs/gardener/gardener-integration-tests.yaml
@@ -21,10 +21,10 @@ presubmits:
         - test-integration
         resources:
           limits:
-            memory: 4Gi
+            memory: 16Gi
           requests:
             cpu: 4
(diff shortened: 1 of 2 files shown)


timebertt:
CLA-Assistant seems to be out of order today 😄 
/override license/cla

PR #210, opened 2022-04-25
Do not inject-effective-version in ci-infra build jobs

oliver-goetz:
/kind bug

The new `--inject-effective-version` make the build jobs in `ci-infra` repo fail. It relies on a VERSION file which is not existing here.
Considering to set the default value of this parameter to `false` instead of true in the future.


timebertt (approved):
> Considering to set the default value of this parameter to `false` instead of true in the future.

Sounds like a good idea 👍 

/lgtm

PR #212, opened 2022-04-25
Allow deployer creating daemonsets and podmonitors

oliver-goetz:
/kind bug

The new kube-prometheus based monitoring needs deployer to be able to create daemonsets (for node-exporter) and podmonitors.


timebertt (approved):
/lgtm

PR #213, opened 2022-04-25
Null routing for KubeAggregatedAPIErrors

oliver-goetz:
/kind bug

We picked up the wrong category to ignore. Enable null-routing for `KubeAggregatedAPIErrors` instead of `KubeAggregatedAPIDown`.


timebertt (approved):
Well..... Good catch!
/lgtm

PR #214, opened 2022-04-25
Change default for inject-effective-version to false

oliver-goetz:
/kind enhancement

Change default values of `--inject-effective-version` in `image-builder` to `false`.
When setting the parameter to `true` image-builder requires a `VERSION` file in the root directory of the Github repository. This is not always the case, so `false` is less error prone setting.

As a result, `--inject-effective-version` must be set to `true` for all gardener build prow jobs.

### DIFF
--- a/config/jobs/gardener/gardener-build-dev-images.yaml
+++ b/config/jobs/gardener/gardener-build-dev-images.yaml
@@ -31,6 +31,7 @@ postsubmits:
         - --target=gardener-extension-provider-local
         - --add-version-tag=true
         - --add-version-sha-tag=true
+        - --inject-effective-version=true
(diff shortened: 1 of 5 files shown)


timebertt (commented):
You could drop `--inject-effective-version=false` from the two jobs specifying it, if you want

oliver-goetz:
I planned to clean this up later, that the build jobs do not fail, until autobumper pushed the new version of image-builder

Am 26.04.2022 um 08:15 schrieb Tim Ebert ***@***.***>:

﻿

@timebertt commented on this pull request.

You could drop --inject-effective-version=false from the two jobs specifying it, if you want

oliver-goetz:
Replying by email using Outlook does look a bit weird 😅

timebertt:
Got it, thanks.
/lgtm
/approve

PR #215, opened 2022-04-25
Support variants in image-builder

oliver-goetz:
/kind bug

This PR simplifies the update of golang-test image.

The current setup is broken. At the moment the golang-version is set in the prow job definition of the image build job while changes to this file are used as a trigger for the prow job to run. This leads to two problems:
1. The prow job is triggered before the config is updated. Thus, the old image version is built when the version is updated. The job has to be triggered manually for another time to build the right image
2. Updating the prow job definitions in the same PR, leads to failing jobs, because initially a build of the wrong version is started

This is the new approach 
- The golang version of test image (and other ci-infra images) is set in Dockerfile only (as default value of `GOLANG_VERSION` build argument
- `golang-test` images are tagged using a scheme which is compatible with the autobumber (v<date>-<git sha>)
- Prow jobs are updated by autobumper when the build was successful and the image pushed.

Advantages:
- This approach builds the right version
- It has a better resilience. The prow jobs are updated only when the build was successful

With this config `golang-test` image is not tagged with the golang version. However, this is not a big deal as there is only one active golang version for tests and the prow jobs are updated automatically.


timebertt (commented):
> With this config golang-test image is not tagged with the golang version. However, this is not a big deal as there is only one active golang version for tests and the prow jobs are updated automatically.

This might be true today, but this might not always be the case.
Is there a way to support different go versions per job with this approach?
I can imagine we will soon need a test job with go v1.18 for master and v1.17 for the release branches in g/g.

oliver-goetz:
We could include the version into the image name 1.18 like gelang-test-1-18 and not in the tag. 
AFAIK autobumper only supports the v{date}-{sha} tag

timebertt:
test-infra seems to append some stuff to tags, maybe this could be a good idea for us:
https://github.com/kubernetes/test-infra/blob/098435238f043dd05ac613e27f004dac2a346555/config/jobs/kubernetes/sig-release/release-branch-jobs/1.21.yaml#L26

oliver-goetz:
oh yes, extending the tag the beginning or the ending does work with the [regexp](https://github.com/kubernetes/test-infra/blob/098435238f043dd05ac613e27f004dac2a346555/experiment/image-bumper/bumper/bumper.go#L32)
We could introduce a new tagging variant for `<prefix>-v<date>-<sha>` and `v<date>-<sha>-<suffix>`

timebertt:
That sounds good. I like the prefix variant, but maybe we can support both :)

oliver-goetz:
I'll update the PR and setup an example with two versions (golang 1.17 & 1.18)

oliver-goetz:
/retitle Support variants in image-builder

@timebertt Ready for review

There are some test prow jobs for the different jobs we have:
- [new variants for golang-test](https://prow.gardener.cloud/view/gs/gardener-prow/logs/post-ci-infra-build-golang-test-image/1519627044592291840)
- [ci-infra images](https://prow.gardener.cloud/view/gs/gardener-prow/logs/post-ci-infra-build-images/1519626696154681344) (for verification that the existing scenarios are still working)
- [gardener images](https://prow.gardener.cloud/view/gs/gardener-prow/logs/post-gardener-build-dev-images/1519627681543491584) (for verification that the existing scenarios are still working)

prow/cmd/image-builder/buildcontroller.go
```diff
@@ -253,110 +266,50 @@ func (r *buildReconciler) createPVC(ctx context.Context, ibPod *corev1.Pod) (*co
...
-		// Configure the build pod with PVC, node assignment and controller reference
-		r.assignPVC(&pod)
-		r.setNodeAssignment(ibPod, &pod)
-		err = controllerutil.SetControllerReference(ibPod, &pod, r.scheme)
-		if err != nil {
-			return errors.Wrap(err, "set controller reference")
-		}
+			// Set build group
+			var buildGroup string
+			if i == 0 && r.options.cacheRegistry != "" {
```
timebertt, line 301: In `startNextBuildPods` you assume that `r.buildPods` is sorted by build group (i.e. `clonerefs`, then all `createCache` and then all `parallel`).
With multiple targets and build variants, this is no longer the case, right?

prow/cmd/image-builder/buildcontroller.go
```diff
@@ -253,110 +266,50 @@ func (r *buildReconciler) createPVC(ctx context.Context, ibPod *corev1.Pod) (*co
...
-		} else {
-			buildGroup = "parallelBuild"
-		}
+	// Next pods build the targets for the variants
+	for _, variant := range variants {
+		for i, target := range r.options.targets.Strings() {
 
-		// Append build container to build pod
-		pod.Spec.Containers = append(pod.Spec.Containers, kanikoContainer)
+			pod, err := r.definePodForTarget(ibPod, target, variant.name, variant.buildArgs)
```
timebertt, line 294: Pass the entire `variant`?
```suggestion
			pod, err := r.definePodForTarget(ibPod, target, variant)
```

prow/cmd/image-builder/buildcontroller.go
```diff
@@ -495,6 +614,29 @@ func (r *buildReconciler) defineDestinations(target string) ([]string, error) {
...
+
+	destination = fmt.Sprintf("--destination=%s/%s:%s", r.options.registry, target, variant)
+	destinations = append(destinations, destination)
+
+	return destinations, nil
+}
+
+func (r *buildReconciler) validateBaseSHA() error {
+	if len(r.options.headSHA) < 7 {
+		return fmt.Errorf("baseSHA %v is it a correct SHA", r.options.headSHA)
```
timebertt, line 635:
```suggestion
		return fmt.Errorf("headSHA %v is not a correct SHA", r.options.headSHA)
```

prow/cmd/image-builder/buildcontroller.go
```diff
@@ -495,6 +614,29 @@ func (r *buildReconciler) defineDestinations(target string) ([]string, error) {
...
+	destination := fmt.Sprintf("--destination=%s/%s:%s", r.options.registry, target, tag)
+	destinations = append(destinations, destination)
+
+	destination = fmt.Sprintf("--destination=%s/%s:%s", r.options.registry, target, variant)
+	destinations = append(destinations, destination)
+
+	return destinations, nil
+}
+
+func (r *buildReconciler) validateBaseSHA() error {
```
timebertt, line 633:
```suggestion
func (r *buildReconciler) validateHeadSHA() error {
```

config/jobs/ci-infra/build-golang-test-image.yaml
```diff
@@ -2,7 +2,7 @@ postsubmits:
   gardener/ci-infra:
   - name: post-ci-infra-build-golang-test-image
     cluster: gardener-prow-trusted
-    run_if_changed: '^Dockerfile|^Makefile|^config\/jobs\/ci-infra\/build-golang-test-image\.yaml'
+    run_if_changed: '^Dockerfile'
```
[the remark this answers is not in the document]
oliver-goetz: Oh I missed this 😅

prow/cmd/image-builder/main.go
```diff
@@ -96,12 +109,16 @@ func gatherOptions() options {
 	fs.StringVar(&o.dockerfile, "dockerfile", "Dockerfile", "path to dockerfile to be built")
 	fs.Var(&o.targets, "target", "target of dockerfile to be built")
 	fs.Var(&o.kanikoArgs, "kaniko-arg", "kaniko-arg for the build")
+	fs.StringVar(&o.buildDefintion, "build-definition", "", "(optional) folder of github repository which includes the build definition (dockerfile, variants")
```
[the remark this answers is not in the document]
oliver-goetz: done 😄

prow/cmd/image-builder/buildcontroller.go
```diff
@@ -434,6 +387,154 @@ func (r *buildReconciler) defineCloneRefsPod(ibPod *corev1.Pod) (corev1.Pod, err
...
+
+	fileContent, err := ioutil.ReadFile(path.Join(r.options.buildDefintion, variantsFile))
+	if os.IsNotExist(err) {
+		return nil, fmt.Errorf("%s not found", variantsFile)
+	} else if err != nil {
+		return nil, errors.Wrapf(err, "failed to load %s", variantsFile)
+	}
+
+	variants := struct {
+		Variants variants `json:"variants"`
```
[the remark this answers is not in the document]
oliver-goetz: 👍

prow/cmd/image-builder/buildcontroller.go
oliver-goetz: done 👍

prow/cmd/image-builder/buildcontroller.go
oliver-goetz: done 😃

prow/cmd/image-builder/buildcontroller.go
```diff
@@ -434,6 +387,154 @@ func (r *buildReconciler) defineCloneRefsPod(ibPod *corev1.Pod) (corev1.Pod, err
...
+		for v := range variants.Variants {
+			if v != r.options.buildVariant {
+				delete(variants.Variants, v)
+			}
+		}
+	}
+
+	var buildVariants []buildVariant
+	for variant, buildArgs := range variants.Variants {
+		v := variant
```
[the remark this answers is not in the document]
oliver-goetz: 👍

prow/cmd/image-builder/buildcontroller.go
oliver-goetz: When cache is enabled and there are multiple variants the groups and their sequence will look like this
- **1** "clonerefs" 
- **2** "createCache-variant1"
- **3** "parallelBuild" (for variant 1)
- **3** "parallelBuild" (for variant 1)
- **3**  ... (for variant 1)
- **4** "createCache-variant2"
- **5** "parallelBuild" (for variant 2)
- **5** ... (for variant 1)

The slice is not sorted after it was created, so it will always build one variant after another. Thus, variant 2 could use cached layers created by variant 1.

When you disable cache everything except the cloning will run in parallel
- **1** "clonerefs"
- **2** "parallelBuild" (for variant 1)
- **2** ... (for variant 1)
- **2** "parallelBuild" (for variant 2)
- **2** ... (for variant 2)

The reason for adding variant name to the build group is when have multiple variants of a single stage Dockerfile and you enable cache explicitly
- **1** "clonerefs" 
- **2** "createCache-variant1"
- **3** "createCache-variant2"
In this case variant 2 would be able to consume the cache created by variant 1 as in the multi stage case.

oliver-goetz:
@timebertt reworked based on your feedback 😄 

I tagged `golang-test:1.17-v20220425-f5c338c` manually that the first jobs after merging do not fail

prow/cmd/image-builder/buildcontroller.go
timebertt: Alrighty. Then this seems to be what we want. I wasn't sure, if this is by intent, but makes sense :)

prow/cmd/image-builder/main.go
```diff
@@ -109,16 +106,16 @@ func gatherOptions() options {
 	fs.StringVar(&o.dockerfile, "dockerfile", "Dockerfile", "path to dockerfile to be built")
 	fs.Var(&o.targets, "target", "target of dockerfile to be built")
 	fs.Var(&o.kanikoArgs, "kaniko-arg", "kaniko-arg for the build")
-	fs.StringVar(&o.buildDefintion, "build-definition", "", "(optional) folder of github repository which includes the build definition (dockerfile, variants")
-	fs.StringVar(&o.buildVariant, "build-variant", "", "variant of build-definition which should be built. Builds all variants if empty")
-	fs.StringVar(&o.registry, "registry", "", "container registry where build artifacts are being pushed. Cache is disabled for empty value")
-	fs.StringVar(&o.cacheRegistry, "cache-registry", "", "container registry where cache artifacts are being pushed")
+	fs.StringVar(&o.context, "context", "", "(optional) context in github repository which includes the build definition (dockerfile, variants")
```
timebertt, line 109:
```suggestion
	fs.StringVar(&o.context, "context", "", "(optional) context in github repository which includes the build definition (dockerfile, variants)")
```

timebertt (commented):
Nice, looks great!
Do you want to add some test cases for image variants?

oliver-goetz:
> Nice, looks great! Do you want to add some test cases for image variants?

I added one and extended another one

timebertt (approved):
Thank you!
/lgtm

oliver-goetz:
We could include the version into the image name 1.18 like gelang-test-1-18 and not in the tag.
AFAIK autobumper only supports the v{date}-{sha} tag

Am 26.04.2022 um 08:21 schrieb Tim Ebert ***@***.***>:

﻿

@timebertt commented on this pull request.

With this config golang-test image is not tagged with the golang version. However, this is not a big deal as there is only one active golang version for tests and the prow jobs are updated automatically.

This might be true today, but this might not always be the case.
Is there a way to support different go versions per job with this approach?
I can imagine we will soon need a test job with go v1.18 for master and v1.17 for the release branches in g/g.

PR #217, opened 2022-04-26
Use latest grafana version with an Apache2.0 license

oliver-goetz:
/kind cleanup

Switch to `grafana:7.5.15` which is currently its latest version licensed with Apache2.0.

### DIFF
--- a/config/prow/cluster/monitoring/base-prow/grafana-deployment.yaml
+++ b/config/prow/cluster/monitoring/base-prow/grafana-deployment.yaml
@@ -8,7 +8,6 @@ spec:
     spec:
       containers:
       - name: grafana
-        image: grafana/grafana:7.5.15
         env:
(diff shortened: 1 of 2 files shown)


config/prow/cluster/monitoring/base-prow/grafana-deployment.yaml
```diff
@@ -8,6 +8,7 @@ spec:
     spec:
       containers:
       - name: grafana
+        image: grafana/grafana:7.5.15
```
timebertt, line 11: If you want, you can use the `images` field in kustomize to overwrite the image ref in a central place :)
https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/images/

oliver-goetz: it works 😄

timebertt (approved):
/lgtm

PR #228, opened 2022-05-02
Use golang-test:1.17-...again

oliver-goetz:
/bug

Unfortunately autobumper does not handle bumps with a prefix correctly.
Manual downgrade to golang 1.17


oliver-goetz:
The PR now includes a proper fix. Image-builder creates a tag with a suffix instead of a prefix now.

timebertt (commented):
Thank you!

timebertt (approved):
/lgtm

PR #236, opened 2022-05-06
Correct command help URL in welcome message

oliver-goetz:
/kind bug

Fixing a minor error in welcome message template.

Btw. I looked for other URLs pointing to `kubernetes/test-infra` repository. Unfortunately, the other URLs used by prow plugins (like in the **Details** section of prow comments) are hardcoded.


timebertt (approved):
Thanks!
/lgtm

PR #245, opened 2022-05-13
Null routing for KubeCPUOvercommit and KubeClientCertificateExpiration alerts

oliver-goetz:
/kind enhancement

This PR creates two additional null routings:
- `KubeCPUOvercommit`: image builds create peak demands which makes the event fire. On a node failure running build jobs will fail anyhow and need to be restarted, Thus, we don't want to provide spare nodes for these jobs.
- `KubeClientCertificateExpiration`: this alert fires anytime someone connects with `gardenctl` to the cluster, so it is rather annoying than helpful

### DIFF
--- a/config/prow/cluster/monitoring/base-prow/alertmanager/config.yaml
+++ b/config/prow/cluster/monitoring/base-prow/alertmanager/config.yaml
@@ -15,7 +15,7 @@ route:
   # Metrics not visible in shoot cluster
   - receiver: 'null-receiver'
     matchers:
-      - alertname=~"KubeSchedulerDown|KubeControllerManagerDown|KubeAggregatedAPIErrors|KubeAggregatedAPIDown"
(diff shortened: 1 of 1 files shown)


config/prow/cluster/monitoring/base-prow/alertmanager/config.yaml
```diff
@@ -15,7 +15,7 @@ route:
   # Metrics not visible in shoot cluster
   - receiver: 'null-receiver'
     matchers:
-      - alertname=~"KubeSchedulerDown|KubeControllerManagerDown|KubeAggregatedAPIErrors|KubeAggregatedAPIDown"
+      - alertname=~"KubeSchedulerDown|KubeControllerManagerDown|KubeAggregatedAPIErrors|KubeAggregatedAPIDown|KubeClientCertificateExpiration|KubeCPUOvercommit"
```
timebertt, line 18:
```suggestion
      - alertname=~"KubeSchedulerDown|KubeControllerManagerDown|KubeAggregatedAPIErrors|KubeAggregatedAPIDown|KubeClientCertificateExpiration|KubeCPUOvercommit|KubeMemoryOvercommit"
```

timebertt (approved):

PR #246, opened 2022-05-13
Update golang versions

oliver-goetz:
/kind enhancement

Go versions 1.17.10 & 1.18.2 are available

### DIFF
--- a/config/jobs/ci-infra/ci-infra-periodics.yaml
+++ b/config/jobs/ci-infra/ci-infra-periodics.yaml
@@ -273,7 +273,7 @@ periodics:
     description: Runs go tests for prow developments in ci-infra 
   spec:
     containers:
-    - image: golang:1.17.10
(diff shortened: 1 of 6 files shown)


config/jobs/ci-infra/ci-infra-periodics.yaml
```diff
@@ -273,7 +273,7 @@ periodics:
     description: Runs go tests for prow developments in ci-infra 
   spec:
     containers:
-    - image: golang:1.17.9
+    - image: golang:1.17.10
```
timebertt, line 276: We could use the `golang-test` image with a moving tag for ci-infra jobs as well. WDYT?
Then we wouldn't need to manually adapt this.

oliver-goetz: good idea, done 😄

PR #255, opened 2022-05-19
Use different test jobs for go 1.17 and 1.18 in release branches

oliver-goetz:
/kind enhancement

Since `gardener v1.47` we are using go 1.18.
Thus, release branches v1.47 and later should be tested with the 1.18 version of golang-test whereas previous version should continue using 1.17


shafeeqes (approved):
/lgtm

PR #263, opened 2022-05-30
Update nginx to v1.2.0 / cleanup finished admission webhook jobs with kustomize

oliver-goetz:
/kind enhancement

Update to version v1.2.0 of nginx. 
Additionally, add `ttlSecondsAfterFinished: 0` to admission webhook jobs for cleaning them up by using kustomize. Those jobs had to be cleaned up manually before. Using kustomize for every deployment is on our agenda anyway.
The helm chart is doing the same ([1](https://github.com/kubernetes/ingress-nginx/blob/08848d69e0c83992c89da18e70ea708752f21d7a/charts/ingress-nginx/templates/admission-webhooks/job-patch/job-createSecret.yaml#L20-L23) & [2](https://github.com/kubernetes/ingress-nginx/blob/08848d69e0c83992c89da18e70ea708752f21d7a/charts/ingress-nginx/templates/admission-webhooks/job-patch/job-patchWebhook.yaml#L20-L23)), but evaluating `APIVersions` does not work when using `helm template`.


timebertt (approved):
Nice one!
/lgtm

PR #264, opened 2022-05-30
Deploy prow components using kustomize

oliver-goetz:
/kind enhancement

Prow components themselves are the last pieces we do not deploy with kustomize yet. This PR changes that.


config/prow/cluster/build/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- test-pods_namespace.yaml
+- cpu-limit-range_limitrange.yaml
+- mem-limit-range_limitrange.yaml
+- gce-ssd_storageclass.yaml
```
timebertt, line 8: These resources are placed on both clusters. Using kustomize, we could easily share the manifests now :)

timebertt (commented):
Nice, love it!

config/prow/cluster/build/kustomization.yaml
oliver-goetz: good point 😄 
I created the overlay for these settings and renamed the folder from `build` to `base`. It would sound weird if the trusted cluster inherits settings from the build cluster. The build cluster having just a `base` configuration sounds ok from my point of view

timebertt (approved):
Nice!
/lgtm

PR #267, opened 2022-06-02
Update kubectl image to v1.22.10

oliver-goetz:
/kind enhancement

Our kubectl image is outdated. Updating to v1.22.10 which is the latest version fitting to the minor version skew of +/-1 to our cluster.


timebertt (approved):
/lgtm

BTW: we could also upgrade our clusters to k8s 1.22

PR #275, opened 2022-06-07
Update shoot spec

oliver-goetz:
/kind task

Catching up with cluster reality:
- Prow clusters are upgraded to `Kubernetes v1.22.7`.
- `image-builder` worker group of trusted cluster is now able to scale down to 0
- trusted cluster uses now `n2d` instead of `e2` machines for base worker group too because they offer more CPU power for the same price in GCP region `europe-west1`


timebertt (approved):
Awesome, thank you!
/lgtm

PR #276, opened 2022-06-08
Increase CPU limits

oliver-goetz:
/kind enhancement

There are some alerts about CPU throttling for
- node-exporter
- node-exporter -> kube-rbac-proxy
- nginx-defaultbackend
This PR raises their CPU limits a bit.

There is a new patch version for nginx (v1.2.1) too.


timebertt (approved):
Thank you!
/lgtm

PR #280, opened 2022-06-09
Highlight standard ginkgo error from e2e tests

oliver-goetz:
/kind enhancement

In E2E we always assert `To(Succeed())`.
Ginkgo error messages start with ["Expected success, but got an error"](https://prow.gardener.cloud/view/gs/gardener-prow/pr-logs/pull/gardener_gardener/6090/pull-gardener-e2e-kind/1534949945625284608#1:build-log.txt%3A1931) in this case.


timebertt (approved):
/lgtm

PR #293, opened 2022-06-20
Add git to cherrypicker image

oliver-goetz:
/kind bug

cherrypicker is crash looping because it does not include a git binary.
```
k logs -n prow cherrypicker-9dd5dc76c-p6gcm                                                                                                                                                                                                
{"client":"github","component":"unset","file":"/code/vendor/k8s.io/test-infra/prow/github/client.go:910","func":"k8s.io/test-infra/prow/github.(*client).log","level":"info","msg":"Throttle(0, 0, *)","severity":"info","time":"2022-06-20T19:39:49Z"}
{"component":"unset","error":"exec: \"git\": executable file not found in $PATH","file":"/code/prow/external-plugins/cherrypicker/main.go:111","func":"main.main","level":"fatal","msg":"Error getting Git client.","severity":"fatal","time":"2022-06-20T19:39:49Z"}
``` 

This PR adds `git` to the cherrypicker image. It continues using the alpine image.
Another option could be using `gcr.io/k8s-prow/git` as it seems to be done in [kubernetes/test-infra](https://github.com/kubernetes/test-infra/blob/ae8821c9316a07a76d6334c2b6209db674ec7ed5/.ko.yaml#L38), but I prefer consistent base images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,11 +23,18 @@ RUN apk add --no-cache ca-certificates
 RUN addgroup -g 65532 -S nonroot && adduser -u 65532 -S nonroot -G nonroot
 USER 65532
 
+FROM alpine:3.15.4 AS ssl_git_runner
+# Install SSL ca certificates
+RUN apk add --no-cache ca-certificates git
+# Create nonroot user and group to be used in executable containers
+RUN addgroup -g 65532 -S nonroot && adduser -u 65532 -S nonroot -G nonroot
+USER 65532
+
 # ----------------------
 # Executable containers
 # ----------------------
 
-FROM ssl_runner AS cherrypicker
+FROM ssl_git_runner AS cherrypicker
 LABEL app=cherrypicker
 WORKDIR /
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -23,11 +23,18 @@ RUN apk add --no-cache ca-certificates
 RUN addgroup -g 65532 -S nonroot && adduser -u 65532 -S nonroot -G nonroot
 USER 65532
 
+FROM alpine:3.15.4 AS ssl_git_runner
```
timebertt, line 26: Why don't you inherit from `ssl_runner` and only add git?

timebertt (commented):
Otherwise lgtm

Dockerfile
oliver-goetz: `apk add` does not work anymore after `USER 65532` is set and there is no `sudo` in alpine installed by default.
Having a new image looked cleaner for me comparing to switching `USER` back and forth.

timebertt (approved):
/lgtm

PR #297, opened 2022-06-23
Revert "Update Prow to v20220623-3ef0bc3b81"

timebertt:
Reverts gardener/ci-infra#296
because it seems to break spyglass again.


timebertt:
Spyglass works again after reverting.

PR #315, opened 2022-07-13
Increased CPU requests for gardener e2e tests

oliver-goetz:
/kind enhancement

The idea of https://github.com/gardener/ci-infra/pull/312 that high CPU consumptions of integration and unit tests have been the reason for flakey e2e tests was wrong.
e2e Tests themselves are creating big CPU loads which are not visible using Kubernetes APIs (maybe because of the docker in docker approach), but you can see them on the nodes. With the current CPU requests of 5 there could be 2 e2e tests running on one node at the same time. This leads to incredible loads from 200-500. We should avoid that and ensure that only one e2e test is running on one node.

On the other side integration and unit tests are not on the critical path, so it would be ok if there are throttled. By doing so, we can avoid using an entire node per test.


timebertt (approved):
Thank you!
/lgtm

PR #363, opened 2022-08-25
Adapt gardener release branch jobs to change of golang in release v1.54

oliver-goetz:
/kind bug

Release jobs for Gardener release currently fail because of the switch to golang 1.19 in Gardener release v1.54.
This PR adapts the release jobs that unit and integration tests for <v1.54 continue to run with golang 1.18, but those for v1.54 and later are carried out with golang 1.19.


timebertt (approved):
Thank you!
/lgtm

PR #368, opened 2022-08-29
Increase CPU limits of kube-state-metrics

timebertt:

Increase CPU limits of kube-state-metrics to decrease CPU throttling.


oliver-goetz (approved):
/lgtm

A yes, there are quite some alerts in slack, I used to ignore 😅

PR #371, opened 2022-08-31
Update nginx to v1.3.0

oliver-goetz:
/kind enhancement

ingress-nginx v1.3.0 is available: [ref](https://github.com/kubernetes/ingress-nginx/releases/tag/controller-v1.3.0)

Additionally, it fixes permissions of the update script.


timebertt (approved):
/lgtm

PR #372, opened 2022-08-31
Update kube-prometheus to v0.11.0

oliver-goetz:
/kind enhancement

This PR is changing three things
- Reduce `retention` of prometheus to 30 days
  - This should reduce the startup time of prometheus in build cluster when nodes are scaled down
  - There is no real use to save data for a longer period
- Update kube-prometheus to v0.11.0: [ref](https://github.com/prometheus-operator/kube-prometheus/releases/tag/v0.11.0)
- Update grafana to 7.5.16, which is the latest version using an Apache 2.0 license: [ref](https://grafana.com/grafana/download/7.5.16?edition=oss)

### DIFF
--- a/config/prow/cluster/monitoring/base-prow/kustomization.yaml
+++ b/config/prow/cluster/monitoring/base-prow/kustomization.yaml
@@ -9,7 +9,7 @@ bases:
 
 images:
 - name: grafana/grafana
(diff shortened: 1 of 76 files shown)


timebertt (approved):
Thanks!
/lgtm

PR #374, opened 2022-09-01
Fix grafana network policy

oliver-goetz:
/kind bug

Network policy for Grafana introduced with kube-prometheus v0.11.0 does not allow ingress traffic from ingress-nginx.
This PR allows ingress originated from the respective namespace.


config/prow/cluster/monitoring/base-prow/grafana-networkPolicy.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: grafana
+  namespace: monitoring
+spec:
+  ingress:
+  - from:
```
timebertt, line 8: This patch drops the port setting

timebertt:
ref https://github.com/gardener/ci-infra/pull/372

PR #376, opened 2022-09-01
Lower tide's sync period

timebertt:

Lower tide's sync period to `1m`, which is also the default ([ref](https://github.com/kubernetes/test-infra/blob/61ae301c157fd3a97e515132c3fe9f616a43c8b9/prow/config/prow-config-documented.yaml#L1331-L1332)).
This makes tide more responsive.
/kind enhancement


oliver-goetz (approved):
/lgtm

PR #380, opened 2022-09-05
Increase cpu limit of kube-rbac-proxy in prometheus-operator

oliver-goetz:
/kind enhancement

kube-rbac-proxy of prometheus-operator produces many slack warning about cpu-throttling.


timebertt (approved):
Thanks!
/lgtm

PR #381, opened 2022-09-05
Add release milestone check for gardener release v1.55 to tide

oliver-goetz:
/kind enhancement

This PR enables milestone check for gardener release `v1.55` in Tide.

/hold until release milestone has been created in `gardener/gardener`

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -163,7 +163,6 @@ tide:
   queries:
   - repos:
     - gardener/ci-infra
-    - gardener/gardener
     labels:
     - lgtm
     - approved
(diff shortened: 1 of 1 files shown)


config/prow/config.yaml
```diff
@@ -195,6 +195,11 @@ tide:
     - do-not-merge/work-in-progress
     - needs-rebase
     - "cla: no"
+  - repos:
+    - gardener/gardener
+    milestone: v1.55
+    includedBranches:
+    - master
```
timebertt, line 202: Can you move this section up, so that it is next to the other g/g config?

I think we need to exclude `master` from the other g/g config. At least, that's what the kubernetes folks do: https://github.com/kubernetes/test-infra/commit/6c829da596d67eff62e0c3885992f6cc34ea251b
Also, we need the same `labels` and `missingLabels` sections. Otherwise, all PRs will be automerged as soon as they are added to the milestone independent of test results 😄

oliver-goetz: I was hoping that Tide considers all the queries for the given repositories. Atm. the configs for `gardener/gardener` and `gardener/ci-infra` are the same. If my assumption is true, we would be able to keep it that way. Otherwise, we to split it up.
I'check this in the Tide codebase.

oliver-goetz: @timebertt I split the config for `gardener/ci-infra` and `gardener/gardener`

timebertt (approved):
Awesome, looks great!
/lgtm

PR #441, opened 2022-10-17
Changes from prow cluster migration

oliver-goetz:
/kind enhancement

There have been some changes and finding during our recent prow cluster migration, which are addressed in this PR.
- Renaming the cluster
- Removing non-existing additional scrape config from prometheus of build cluster
- Moving prow cluster specifications to another place


rfranzke (approved):
/lgtm

PR #442, opened 2022-10-17
Increase memory request of e2e tests

oliver-goetz:
/kind enhancement

This PR increases memory requests of e2e test.
Most importantly is the increase for HA e2e tests to 64Gi, because they seem to request much more memory than the standard e2e tests.
Additionally, memory limits for e2e tests are removed. They are not enforced anyway, because the memory is not requested by the pod of the prow job itself, but the docker container it is starting.


rfranzke (commented):
/lgtm
/approve

PR #444, opened 2022-10-17
Remove check-vulnerabilities jobs from gardener/gardener

oliver-goetz:
/kind enhancement

`govulncheck` reports vulnerabilities from imported libraries we cannot influence, but triage only. It does not have an option to exclude vulnerabilities, so it always fails and nobody really checks its results.
This PR removes prow jobs running `make check-vulnerabilities` which starts `govulncheck` for gardener/gardener repository.


rfranzke (approved):
/lgtm

PR #487, opened 2022-11-10
Increase cpu limit of kube-rbac-proxy-main in kube-state-metrics deployment

oliver-goetz:
/kind enhancement

There are many CPU throttling alerts for kube-rbac-proxy-main in kube-state-metrics deployment especially in build cluster. We are not lacking of CPU power in general, so this PR increases the limit.

Fixes #


timebertt (approved):
/lgtm

PR #528, opened 2022-12-16
Bind prometheus pods to system-component nodes

oliver-goetz:
/kind enhancement

This PR binds various `prometheus` pods to `worker.gardener.cloud/system-components` nodes, that they do not run on the same nodes as e2e tests. 
e2e tests can create such high loads on a node, that prometheus is not able to scrape properly meanwhile.

Fixes #


rfranzke (commented):
/lgtm
/approve

PR #623, opened 2023-03-15
Create docker configuration for dind to enable using nodes with cgroupv2

oliver-goetz:
/kind enhancement

This PR defines a docker `daemon.json` which enables us running docker in docker e2e tests on nodes with `cgroupv2` and mounts it into the respective pods.
It works on nodes using `cgroupv1` too, so we can already apply it in advance.

Fixes #

### DIFF
--- a/config/prow/cluster/base/dind-docker-config_configmap.yaml
+++ b/config/prow/cluster/base/dind-docker-config_configmap.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: dind-docker-config
+  namespace: test-pods
+data:
+  daemon.json: |-
+    {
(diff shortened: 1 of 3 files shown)


config/prow/cluster/base/dind-docker-config_configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: dind-docker-config
+  namespace: test-pods
+data:
+  daemon.json: |-
+    {
+      "cgroup-parent": "prowparent.slice"
```
timebertt, line 9: Does this cgroup need to exist beforehand?
Is this created explicitly somewhere?

timebertt (commented):
I have a question. But generally,
/lgtm

Issue #619, opened 2023-03-10
Copy container images from docker hub to gardener GCR


timebertt:
/kind enhancement

**What would you like to be added**:

Let's copy all docker hub images that the gardener project consumes to our gardener GCR and consume them from there.

To support this effort with automation, let's add a CI job that copies a configured set of container images based on a simple configuration file:

```yaml
images:
- source: kubernetesui/dashboard
  destination: eu.gcr.io/gardener-project/3rd/kubernetesui/dashboard
  tags:
  - v2.2.0
  - v2.4.0
  - v2.5.1
- source: envoyproxy/envoy-distroless
  destination: eu.gcr.io/gardener-project/3rd/envoyproxy/envoy-distroless
  tags:
  - v1.24.1
```

For each container image, the configuration file specifies one or multiple tags that should be copied from the given source repository to the given destination repository.
The copied images should include all platforms/architectures supported by the source images.

**Why is this needed**:

The gardener project consumes multiple images from docker hub. This leads to problems because

- docker hub enforces strict [rate limits for image pulls](https://docs.docker.com/docker-hub/download-rate-limit/)
- docker hub doesn't [support image pulls over IPv6](https://www.docker.com/blog/beta-ipv6-support-on-docker-hub-registry/) (ref https://github.com/gardener/gardener/issues/7051)

With our gardener GCR, we already have the necessary infrastructure to solve both problems. GCR has no rate limit and supports IPv6.

Previously, some container images were already copied to the gardener GCR manually.
This is not enough for switching all container images over to our GCR because

- push access to the GCR is limited: we don't want to bother individuals with the task of copying container images, e.g. when versions change or new images should be used
- manual actions are not reviewed or audited where development is happening (GitHub)
- manual actions are error-prone: e.g., a simple `docker pull` + `docker tag` + `docker push` only includes the architecture of the used machine

Instead, we want to have a structured and automated way of copying all necessary container images based on a central inventory that can be updated in a self-service PR workflow.

**Alternatives considered**:

We can't use the [GCP Docker Hub Mirror](https://cloud.google.com/container-registry/docs/pulling-cached-images) (`mirror.gcr.io`) because it is not a pull-through cache but only caches frequently access images, see the documentation (ref https://github.com/gardener/gardener/pull/7568#discussion_r1120587079):

> Container Registry adds frequently requested images to the cache so they are available for future requests.
> [...]
[...]

PR #636, opened 2023-03-22
Feature/copy images

robinschneider:
/kind enhancement

This PR is the next step in #619 
It adds all DockerHub images from these files to copy them to GCR

https://github.com/gardener/gardener/blob/master/charts/images.yaml 
https://github.com/gardener/gardener/blob/master/pkg/provider-local/imagevector/images.yaml 
https://github.com/gardener/gardener-extension-networking-calico/blob/master/charts/images.yaml

### DIFF
--- a/config/images/README.md
+++ b/config/images/README.md
@@ -33,6 +33,12 @@ It is required to have yq and crane installed.
 brew install yq
 brew install crane
 ```
+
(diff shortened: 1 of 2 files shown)


timebertt (approved):
/lgtm
/ok-to-test

PR #641, opened 2023-03-22
copy-images: switch back to sh

timebertt:
/kind bug

My bad, alpine doesn't have bash but only sh. Shellcheck tells me the `pipefail` option is not defined in POSIX sh, so drop it.

Follow-up to #640 


oliver-goetz:
/approve
/lgtm
🤞

PR #648, opened 2023-03-24
Add new versions and manila-csi-plugin for openstack provider

MartinWeindel:
Added new images to images.yaml from the following repositories:

* gardener-extension-provider-openstack

/kind enhancement

What this PR does / why we need it:
This adds the last missing container images to Gardener GCR that are not yet copied from DockerHub.

related to
- https://github.com/gardener/gardener-extension-provider-openstack/pull/588
- https://github.com/gardener/gardener-extension-provider-openstack/pull/572

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -99,9 +99,17 @@ images:
   - v1.22.2
   - v1.23.4
   - v1.24.5
+  - v1.24.6
   - v1.25.3
+  - v1.25.5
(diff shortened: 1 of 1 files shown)


MartinWeindel:
not needed anymore as openstack-cloud-controller-manager and cinder-csi-plugin images are pushed to GCR, see https://github.com/gardener/gardener-extension-provider-openstack/pull/593

PR #665, opened 2023-03-30
Update alpine to 3.15.8

ScheererJ:
/kind enhancement

Update alpine to 3.15.8

None.

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -21,6 +21,7 @@ images:
 - source: alpine
   destination: eu.gcr.io/gardener-project/3rd/alpine
   tags:
+  - 3.15.4
   - 3.15.8
 - source: fluent/fluent-bit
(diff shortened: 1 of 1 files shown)


config/images/images.yaml
```diff
@@ -21,7 +21,7 @@ images:
 - source: alpine
   destination: eu.gcr.io/gardener-project/3rd/alpine
   tags:
-  - 3.15.4
+  - 3.15.8
```
oliver-goetz, line 24: We should keep the old version too, that we can keep track on the images we already copied for previous versions.

timebertt: If that's the decision we should state this explicitly in `config/images/README.md`

ScheererJ: I kept the old version, but I am also a bit surprised that we want to keep the history in addition to version control (git) also in the file.

PR #667, opened 2023-03-31
Add E2E tests for landscaper

Diaphteiros:
The goal is to make the `/test` command work again (and execute the integration tests) for PRs on the Landscaper.
Currently, the tests should only run if this `/test` command is added manually.


Diaphteiros:
cc @oliver-goetz 
Does it make sense to create a new dashboard group for the Landscaper? I don't know what the implications of that would be.

oliver-goetz (commented):
The dashboard defined in `config/testgrids/config.yaml` defines a dashboard on testgrid like this for cilium: https://testgrid.k8s.io/gardener-extension-networking-cilium

Atm. you do not reference any dashboard in your prow job. Thus, there it is not necessary to define it right now.
However, our periodic prow jobs usually use dashboard. In case you would like to create one (which I would encourage to do), it would make sense, because it easier to check the results of the test runs there.

/kind enhancement

config/jobs/landscaper/landscaper-integration-kind.yaml
```diff
@@ -0,0 +1,38 @@
+presubmits:
```
oliver-goetz, line 1: Please rename the file to `landscaper-e2e-kind.yaml`

config/testgrids/config.yaml
```diff
@@ -8,6 +8,7 @@ dashboard_groups:
     - gardener-dependency-watchdog
     - gardener-extension-networking-cilium
     - gardener-extension-networking-calico
+    - gardener-landscaper
```
oliver-goetz, line 11: If you don't use the dashboard, you have to revert the change. Otherwise, the testgrid test fails.

config/testgrids/config.yaml
```diff
@@ -17,3 +18,4 @@ dashboards:
 - name: gardener-dependency-watchdog
 - name: gardener-extension-networking-cilium
 - name: gardener-extension-networking-calico
+- name: gardener-landscaper
```
oliver-goetz, line 21: same here

oliver-goetz (commented):
There are two more things 😄

PR #749, opened 2023-06-07
update fluent-bit container image to v2.1.4

nickytd:
/kind enhancement

This PR updates fluent-bit container image to the latest released version [v2.1.4](https://github.com/fluent/fluent-bit/releases/tag/v2.1.4)

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -32,6 +32,7 @@ images:
   tags:
   - v2.0.9
   - v2.0.10
+  - v2.1.4
 - source: kubesphere/fluent-operator
   destination: eu.gcr.io/gardener-project/3rd/kubesphere/fluent-operator
   tags:


config/images/images.yaml
```diff
@@ -32,6 +32,7 @@ images:
   tags:
   - v2.0.9
   - v2.0.10
+  - v2.1.4
```
vpnachev, line 35: I was wondering why `kubesphere/fluent-bit` is the source image and not `cr.fluentbit.io/fluent/fluent-bit` or `fluent/fluent-bit` that are referenced in https://docs.fluentbit.io/manual/installation/docker. Then I found 
> A custom Fluent Bit image kubesphere/fluent-bit is required to work with FluentBit Operator for dynamic configuration reloading.

ref: https://github.com/fluent/fluent-operator#fluent-bit

Can you add a short comment that the fluent bit operator requires custom fluent bit image built by the operator community?

nickytd: Nothing to add here. The statement is correct. The reason why that is needed is related to the fact that the `fluent-bit` still lacks `hot` configuration reload. Hence when a new fluent-bit configuration is deployed by the operator that image handles the changes. So if we want to bring new fluent-bit in place we need to fetch the latest released kubesphere/fluent-bit.

PR #776, opened 2023-07-10
Introduce `release-handler` 

oliver-goetz:
/kind enhancement

`release-handler` is meant to automate tasks related to Gardener releases. 
It identifies four cases based on the change of the `VERSION` file.
- `newRelease`: when the version increases from a dev to a final version like `v1.75.0-dev -> v1.75.0`.
- `nextDevCycle`: when the version increases from a final to a higher dev version like `v1.75.0 -> v1.75.1-dev`
- `prepareNewPatchVersion`: when the version increases from one dev version to another with a higher patch version like `v1.75.0-dev -> v1.75.1-dev`
- `prepareNewMajorMinorVersion`: when the version increases from one dev version to another with a higher major or minor version like `v1.75.0-dev -> v1.76.0-dev` 

In this first version in only acts in the `prepareNewMajorMinorVersion` case and creates new release branches for the lower version.
It will be enabled for `gardener/gardener` repository via this PR.

Fixes #

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,3 +59,9 @@ LABEL app=image-builder
 WORKDIR /
 COPY --from=builder /build/image-builder /image-builder
 ENTRYPOINT [ "/image-builder" ]
+
+FROM ssl_git_runner AS release-handler
+LABEL app=release-handler
+WORKDIR /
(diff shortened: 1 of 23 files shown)


rfranzke (commented):
/lgtm
/approve

PR #789, opened 2023-07-14
Introduce `branch-cleaner`

oliver-goetz:
/kind enhancement

This PR introduces `branch-cleaner`.
The `branch-cleaner` scans GitHub repositories for branches of a specific pattern and deletes them.
When the `--keep-branches=N` is set, it keeps the last `N` matching branches sorted in alphabetical order.
When `--ignore-open-prs=false` it checks the branches for open PRs before deletion. If there are any, the branch won't be deleted.
When `--release-branch-mode=true` it checks release branch names for semver versions and searches for an existing Github release tag (`v{major}.{minor}.0`) for this version. Release branches which do not have a version yet are neither counted nor considered for deletion, because they are not released yet. If the branch names to not include a semver version `branch-cleaner` fails in this mode.

In context of `gardener/gardener` it is used to delete release branches which are not supported anymore. It runs every day at 10pm.

Additionally, it cleans up cherry-pick branches from `gardener-ci-robot/gardener` which do not have open PRs anymore.

Fixes #


prow/cmd/branch-cleaner/main.go
```diff
@@ -0,0 +1,111 @@
...
+		return fmt.Errorf("error compiling branch pattern: %w", err)
+	}
+
+	return nil
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.fullRepo, "repository", "", "Repository which should be cleaned up.")
```
timebertt, line 64: Would be great to make this a list of repos, that the tool loops over.
With this, we could have a single periodic job that cleans all cherry-pick branches in the robot's forks

oliver-goetz: good idea, I added support for this 😄

oliver-goetz:
There is a `release-branch-mode` option too now. When enabled it expects a semver version in the release branch name. It only counts the branch if it has a Github release tag. 
This prevents `branch-cleaner` to delete branches because there are branches for an ongoing release validation.

rfranzke (commented):
/lgtm

timebertt (commented):
Awesome!
/approve

PR #794, opened 2023-07-20
Set `retentionSize` to 90GB for prometheus instances

oliver-goetz:
/kind enhancement

We were facing an issue that the PV of a prometheus was full.
This PR sets the [retentionSize](https://github.com/gardener/ci-infra/blob/2ddec27c6f54e47093fede284b773f6caceea9f3/config/prow/cluster/monitoring/base/setup/0prometheusCustomResourceDefinition.yaml#L5466-L5469) for our prometheus instances to 90GB in order to avoid this issue in the future. The PV has 100GB of size, so setting the parameter to 90GB leaves 10 GB for the WAL file.

Fixes #

/cc @ialidzhikov

### DIFF
--- a/config/prow/cluster/monitoring/base-prow/prometheus.yaml
+++ b/config/prow/cluster/monitoring/base-prow/prometheus.yaml
@@ -4,6 +4,7 @@ metadata:
   name: k8s
   namespace: monitoring
 spec:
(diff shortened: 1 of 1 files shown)


ialidzhikov (commented):
/lgtm

PR #813, opened 2023-08-02
Enable dependabot to create update PRs for docker images

oliver-goetz:
/kind enhancement

**What this PR does / why we need it**:
This PR enables `dependabot` to create update PRs for Golang and Alpine docker images similar to how it is done in other gardener repositories ([ref](https://github.com/gardener/gardener-extension-os-gardenlinux/blob/020f32da8659e91bf5877fde7cd26e0259537cd5/.github/dependabot.yaml#L11-L15)).
According to the latest update PRs ([ref](https://github.com/gardener/ci-infra/pull/779), [ref](https://github.com/gardener/ci-infra/pull/780)), there are images in `Dockerfile` and in `images/golang-test` folder. I'm not quite sure if dependabot manages to update the later folder, but let's see 😄 

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
This PR changes a file in `.github` directory. It cannot be merged by `Tide` but needs to be merged manually.

PR #814, opened 2023-08-02
Use dependabot compatible docker image format

oliver-goetz:
/kind bug

The format used in `ci-infra` to specify docker images is most likely not compatible to `dependabot` introduced in https://github.com/gardener/ci-infra/pull/813 yet, because the images are concatenated using different `args`.
This PR removes this concatenation that `dependabot` is (hopefully) able to identify the docker images correctly.

Additionally, it uses `golang:${VERSION}` images instead of `golang:${VERSION}-bullseye` images as base for our test-image, because we use these kinds of base images for building gardener images too.

Fixes #

/cc @ScheererJ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,7 @@
 # ----------------
 # Build container
 # ----------------
-ARG GOLANG_VERSION=1.19.11
-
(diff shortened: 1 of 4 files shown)


ScheererJ (approved):
/lgtm

PR #821, opened 2023-08-03
Add `ok-to-test` label for PRs opened by `dependabot[bot]` automatically

oliver-goetz:
/kind enhancement

@dependabot[bot] is trustworthy, so let's add the `ok-to-test` label automatically to PRs opened by it.
When the tests are starting early, this will save us some waiting time.

Fixes #


config/jobs/common/dependabot.yaml
```diff
@@ -0,0 +1,42 @@
+periodics:
+- name: ci-gardener-dependabot-ok-to-test
```
timebertt, line 2: Instead of using an additional labeling job, you could also add `dependabot` to `plugins.trigger.trusted_apps`: https://github.com/kubernetes/test-infra/blob/9f82a909a9f806681e19febb3b305b7b66c668ed/prow/plugins/plugin-config-documented.yaml#L602-L606

oliver-goetz: awesome 🚀

ScheererJ (approved):
/lgtm

PR #872, opened 2023-08-31
Reduce user's permissions in branch-protection rules

oliver-goetz:
/kind enhancement

We could reduce the permissions of users in the branch-protection rules, because some settings do not have a use case (anymore):
- Remove `allow_deletions`: outdated gardener/gardener branches are deleted by prow now. Other repos don't have the use case, that people with write access only delete branches. Admins and maintainers of a repo will be still able to delete branches.
- Set `enforce_admins` to `true`: Merging PRs with failing required tests is foreseen in the PR review process for Gardener repos. Repository admins and maintainers should follow this process too.

Fixes #

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -137,7 +137,7 @@ branch-protection:
           required_status_checks:
             contexts:
             - license/cla
-          restrictions: # prevent everyone except gardener-prow from pushing/merging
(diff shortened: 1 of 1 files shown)


rfranzke (commented):
/lgtm
/approve

PR #909, opened 2023-09-22
Set fixed debian version in test images

oliver-goetz:
/kind enhancement

This PR sets the `debian` version of test images explicitly to `bookworm` that we don't run into `glibc` issues in case the default base image of `golang` changes.
There is a related PR in gardener repo: https://github.com/gardener/gardener/pull/8542

Fixes #


rfranzke (commented):
/lgtm
/approve

PR #912, opened 2023-09-26
`branch-cleaner` removes branch protection rules of branches it plans to delete

oliver-goetz:
/kind enhancement

After permissions in branch protection rules have been reduced in https://github.com/gardener/ci-infra/pull/872, `branch-cleaner` is no longer able to clean up old release branches ([ref](https://prow.gardener.cloud/view/gs/gardener-prow/logs/ci-gardener-branch-cleaner/1706428598539259904)).
Additionally, there are orphaned branch-protection rules of release branches which have been deleted in the past.
Thus, rather then reverting the change this PR enables `branch-protector` to delete branch protection rules for release branches it deletes. This removes one task for the the release responsible.

Fixes #

/cc @rfranzke

### DIFF
--- a/prow/cmd/branch-cleaner/branchcleaner_test.go
+++ b/prow/cmd/branch-cleaner/branchcleaner_test.go
@@ -39,12 +39,31 @@ func (f *fakeGithubClient) GetBranches(org, repo string, onlyProtected bool) ([]
 	return branches, nil
 }
 
+func (f *fakeGithubClient) GetBranchProtection(org, repo, branch string) (*github.BranchProtection, error) {
+	for _, b := range f.Branches {
+		if b.Name == branch && b.Protected {
+			return &github.BranchProtection{}, nil
+		}
+	}
+
+	return nil, nil
+}
+
 func (f *fakeGithubClient) GetPullRequests(org, repo string) ([]github.PullRequest, error) {
 	var prs []github.PullRequest
 	for _, pr := range f.PullRequests {
 		prs = append(prs, *pr)
 	}
 	return prs, nil
+}
+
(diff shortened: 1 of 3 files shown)


prow/cmd/branch-cleaner/branchcleaner_test.go
```diff
@@ -39,6 +39,16 @@ func (f *fakeGithubClient) GetBranches(org, repo string, onlyProtected bool) ([]
 	return branches, nil
 }
 
+func (f *fakeGithubClient) GetBranchProtection(org, repo, branch string) (*github.BranchProtection, error) {
```
rfranzke, line 42: This never returns an `error`, remove it from the signature?

oliver-goetz: The signature of the method in fake client has to be the same as the one of the regular method ([ref](https://github.com/gardener/ci-infra/blob/ec7d6af0b47b141c12a215440c1390b3fd5911b9/prow/cmd/branch-cleaner/branchcleaner.go#L31), [ref](https://github.com/kubernetes/test-infra/blob/a9938784e8c2501c4ba6518dd5473b95d5dc7580/prow/github/client.go#L168)).
I just added it here because the [upstream Github fake client](https://github.com/kubernetes/test-infra/blob/master/prow/github/fakegithub/fakegithub.go) does not implement this method.

rfranzke: Sorry, I missed that this is the fake implementation 🥲

rfranzke (commented):
/lgtm
/approve

PR #980, opened 2023-10-30
Update ingress-nginx to `v1.94` and enable annotation validations

oliver-goetz:
/kind enhancement

This PR updates ingress-nginx to [v1.94](https://github.com/kubernetes/ingress-nginx/releases/tag/controller-v1.9.4) and enables annotation validations.

Fixes #

### DIFF
--- a/config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
+++ b/config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
@@ -789,7 +789,7 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name:  ingress-nginx-admission
+  name: ingress-nginx-admission
   namespace: ingress-nginx
   annotations:
     "helm.sh/hook": pre-install,pre-upgrade,post-install,post-upgrade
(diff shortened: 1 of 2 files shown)


config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
```diff
@@ -788,16 +789,16 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: ingress-nginx-admission
```
ScheererJ, line 791: This whitespace change seems to be unnecessary.

ScheererJ (commented):
Thanks for the change.

I have a small nit as there seems to be an unnecessary whitespace change, which could be removed. Apart from that it looks good.

config/prow/cluster/ingress-nginx/ingress-nginx-deployment.yaml
oliver-goetz: That's true 😅 
The file is generated from the helm chart. The additional space comes from [here](https://github.com/kubernetes/ingress-nginx/blob/5583f90c7f3525b6267747d6b07541c01218a9ea/charts/ingress-nginx/templates/admission-webhooks/job-patch/role.yaml#L5C8-L5C8) and was introduced lately with [this commit](https://github.com/kubernetes/ingress-nginx/commit/7ce6cc88d8501628fd0fe1fb882579ad7d2a95fe#diff-3d10afa8f48a66b5503389d8b56e9d9741898bc277d6e0c21efa43cfe3c2bf62).

The additional space seems to be already removed on main, so I removed it here too 😄

ScheererJ (approved):
/lgtm

PR #1019, opened 2023-11-20
[cherry-picker] Support copying multiple release-notes to cherry-picked PR

oliver-goetz:
/kind enhancement

Cherry-picker did not support copying multiple release notes section into the cherry-picked PR, but copied the first one only. This PR adds support to copy multiple release-notes to the cherry-picked PR.

Fixes #

### DIFF
--- a/prow/external-plugins/cherrypicker/server.go
+++ b/prow/external-plugins/cherrypicker/server.go
@@ -689,9 +689,6 @@ func releaseNoteFromParentPR(prAuthor, org, repo string, num int, body string) s
 	var output string
 	potentialMatches := releaseNoteRe.FindAllStringSubmatch(body, -1)
 	for i, potentialMatch := range potentialMatches {
-		if potentialMatch == nil {
-			return ""
-		}
 		source := strings.TrimSpace(potentialMatch[4])
 		ref := strings.TrimSpace(potentialMatch[5])
 		if source == "" || ref == "" {
--- a/prow/external-plugins/cherrypicker/server.go
+++ b/prow/external-plugins/cherrypicker/server.go
@@ -686,26 +686,33 @@ func normalize(input string) string {
 // parent PR and formats it as per the PR template so that
 // it can be copied to the cherry-pick PR.
 func releaseNoteFromParentPR(prAuthor, org, repo string, num int, body string) string {
-	potentialMatch := releaseNoteRe.FindStringSubmatch(body)
-	if potentialMatch == nil {
-		return ""
+	var output string
+	potentialMatches := releaseNoteRe.FindAllStringSubmatch(body, -1)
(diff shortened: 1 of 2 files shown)


prow/external-plugins/cherrypicker/server_test.go
```diff
@@ -1172,6 +1172,16 @@ func TestReleaseNoteFromParentPR(t *testing.T) {
 			input:    "```feature developer #999\nUpdate the magic number from 42 to 49\n```",
 			expected: "```feature developer github.com/foo/bar #123 @foo-author\nUpdate the magic number from 42 to 49\n```",
 		},
+		{
+			name:     "test9",
+			input:    "```feature developer\nUpdate the magic number from 42 to 49\n```\n\n```feature operator\nUpdate another magic number\n```",
```
plkokanov, line 1177: Small q: I see that there are two newlines between the release notes. Is this intentional for the test? Usually we only separate multiple release notes with 1 new line.

prow/external-plugins/cherrypicker/server.go
```diff
@@ -686,26 +686,33 @@ func normalize(input string) string {
...
-		strings.TrimSpace(potentialMatch[3]),
-		source,
-		ref,
-		author,
-		strings.TrimSpace(potentialMatch[7]),
-	)
+	var output string
+	potentialMatches := releaseNoteRe.FindAllStringSubmatch(body, -1)
+	for i, potentialMatch := range potentialMatches {
+		if potentialMatch == nil {
```
plkokanov, line 692: Can we still get a `potentialMatch` that is nil here? If there are no matches I guess `potentialMatches` will be nil. As far as I understand it, if `potentialMatches` is not nil then there should be at least one submatch, namely the `potentialMatch[0]` one.

prow/external-plugins/cherrypicker/server.go
```diff
@@ -686,26 +686,33 @@ func normalize(input string) string {
...
+			return ""
```
plkokanov, line 693: If a single `potentialMatch` can be nil, shouldn't we just `continue` here instead of returning?

plkokanov (changes requested):
Thanks for the PR!
I have a few minor comments/questions.

prow/external-plugins/cherrypicker/server_test.go
oliver-goetz: I removed the second `\n` 😄

prow/external-plugins/cherrypicker/server.go
oliver-goetz: You are right `potentialMatches == nil` was a special handling when `FindStringSubmatch` was used.

PR #1028, opened 2023-11-29
Revert version of `gcsweb`

oliver-goetz:
/kind bug

This PR reverts the version of `gcsweb` to `gcr.io/k8s-prow/gcsweb:v20231121-4e39ac27ea`.
The current version introduced 
 breaks our setup because of an upstream regression introduced in https://github.com/kubernetes/test-infra/pull/31241.
With this change `gcsweb` is not accessible anymore but shows the following error message.
```
Error: error while processing object: googleapi: Error 403: Insufficient Permission, insufficientPermissions
```

Fixes #

/cc @timebertt


acumino (commented):
/lgtm
/approve

PR #1056, opened 2023-12-13
Never initialize `logrus.StandardLogger().Formatter` in `main()` functions

oliver-goetz:
/kind bug

Initializing `logrus.StandardLogger().Formatter` can be dangerous in case [prow secret agent ](https://github.com/kubernetes/test-infra/blob/master/prow/config/secret/agent.go) is used, because this may overwrite [the censoring logger](https://github.com/kubernetes/test-infra/blob/ce9ca273df909cf5c3acbbd4e3958f406951e691/prow/config/secret/agent.go#L39) which is setup in its `init()` function. This may result in secrets like tokens being logged.

Thus, this PR removes all initializations `logrus.StandardLogger().Formatter` in `main()` functions regardless prow secret manager is used or not.

Fixes #

### DIFF
--- a/prow/cmd/branch-cleaner/main.go
+++ b/prow/cmd/branch-cleaner/main.go
@@ -83,7 +83,7 @@ func gatherOptions() options {
 func main() {
 	logrusutil.Init(&logrusutil.DefaultFieldsFormatter{
 		PrintLineNumber:  true,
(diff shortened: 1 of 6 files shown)


timuthy (approved):
/lgtm

PR #1066, opened 2023-12-19
Prepare switch from GCR to Artifact-Registry

ccwienk:
/kind technical-debt

**What this PR does / why we need it**

Prepare switching of Gardener-Pipelines to publish images from GCR to Artifact-Registry (motivated by [deprecation of GCR](https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr)).

- configure 3rd-party images to be published to AR instead of GCR
- configure images used by/for prow to be published to AR instead of GCR

**Special notes for your reviewer**

### DIFF
--- a/config/images/README.md
+++ b/config/images/README.md
@@ -11,13 +11,13 @@ The images needed are all listed inside the `images.yaml` with the following sch
 ```yaml
 images:
 - source: kubernetesui/dashboard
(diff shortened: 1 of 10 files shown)


oliver-goetz:
awesome 🚀 

/lgtm
/approve

PR #1108, opened 2024-01-23
Fix `ulimit` error in `krte` image when starting docker service with docker-ce 25.0.0

oliver-goetz:
/kind bug

Starting docker in `krte` image fails with docker-ce 25.0.0 with an ulimit error ([ref](https://prow.gardener.cloud/view/gs/gardener-prow/pr-logs/pull/gardener_gardener/9048/pull-gardener-e2e-kind-upgrade/1749725781455540224)).
This PR applies removes the `ulimit` command from `/etc/init.d/docker` as suggested [here](https://github.com/docker/cli/issues/4807#issuecomment-1903950217).

Fixes #


acumino (commented):
/lgtm
/approve

PR #1128, opened 2024-02-05
Copy registry:3.0.0-alpha.1

ialidzhikov:
/kind enhancement

We decided to evaluate the upgrade of the registry image to `registry:3.0.0-alpha.1`. Yes, it is an alpha image but it would unblock us with 2 tasks:
- the monitoring task: `3.0.0-alpha.1` exposes metrics of the proxy in prometheus native format
- configurable TTL: currently one of the limitations of the proxy is that the images TTL is hard-coded to 7 days. This means that a blob lives 7 days in the cache and after that it is always deleted. This reduces the effectiveness of the proxy. Bigger TTL values can make the proxy more effective. We want to expose the TTL field in the registry API.

Part of https://github.com/gardener/gardener-extension-registry-cache/issues/3

N/A

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -170,6 +170,7 @@ images:
   - 2.8.1
   - 2.8.2
   - 2.8.3
+  - 3.0.0-alpha.1
 # gardener-extension-registry-cache/test/e2e
 - source: nginx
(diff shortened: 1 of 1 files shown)


dimitar-kostadinov (commented):
/lgtm

PR #1333, opened 2024-03-13
Add v2.7.1 image of aws load balancer controller 

namewasntfound:
**Please select the kind of this pull request, e.g.**:
/kind enhancement

Related to https://github.com/gardener/gardener-extension-provider-aws/pull/886

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -133,6 +133,7 @@ images:
   tags:
   - v2.6.1
   - v2.7.0
+  - v2.7.1
 - source: public.ecr.aws/ebs-csi-driver/volume-modifier-for-k8s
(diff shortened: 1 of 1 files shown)


config/images/images.yaml
```diff
@@ -133,6 +133,7 @@ images:
   tags:
   - v2.6.1
   - v2.7.0
+  - v2.7.1
```
AndreasBurger, line 136:
```suggestion
  - v2.7.1
```
We do not actually use the `v2.7.1` one, so I think it can be removed.

namewasntfound: we can remove the v2.7.0?

AndreasBurger (approved):

PR #1452, opened 2024-04-08
Increase volume size of `renovate` cache `pvc`

oliver-goetz:
/kind enhancement

The cache disk of `renovate` is constantly close to its inode limit. If you don't want to format a disk with custom settings and attach it manually to Kubernetes (which I don't want), the number of inodes depends on the disk size.
Thus, this PR increases the size of renovate PVC that we don't hit the inode limit.

Fixes #


ScheererJ (approved):
/lgtm
/approve

PR #1900, opened 2024-06-17
fix fluent-operator source

nickytd:
This PR fixes the source of fluent-operator. The correct one shall be `ghcr.io/fluent/fluent-operator/fluent-operator`


oliver-goetz (approved):
Thanks 🙂 
/lgtm

PR #1978, opened 2024-07-02
Allow version suffixes in the container copy logic

ScheererJ:
Allow version suffixes in the container copy logic.

For example, `kindest/local-path-provisioner` has a version `v0.0.22-kind.0`. The renovate templating uses the following:

```
               "currentVersion": "v0.0.22",
               "currentValue": "v0.0.22-kind.0",
```

Therefore, we should use `currentValue` instead of `currentVersion` in the post update step.

Fixes #

/kind enhancement
/cc @oliver-goetz


oliver-goetz (approved):
🚀

PR #2507, opened 2024-10-09
Increase memory requests/limits for gardener/gardener unit tests

oliver-goetz:
/kind enhancement

There are many OOMKilled Pods for Gardener unit tests at the moment. Apparently, `make check-generate` has a memory leak and consumes over 20GB of RAM.
This PR increases memory requests and limits, that we can continue to work 😅 

Fixes #

/cc @ScheererJ


ScheererJ (approved):
/lgtm
/approve

PR #2536, opened 2024-10-16
Add jobs and config for gardener-extension-provider-equinix-metal

RiRa12621:
/kind enhancement
/kind test

We would like to use prow for gardener-extension-provider-equinix-metal related tests as well to unifi the experience


RiRa12621:
/assign @acumino 
obey the bots

PR #2617, opened 2024-10-31
Deduplicate `labels.yaml`

timebertt:
/kind cleanup

This PR deduplicates the `labels.yaml` file as much as possible using yaml anchors.

There are three different sets of labels:
1. List of labels that are used if the repository is fully onboarded to prow.
2. List of labels that are used if the repository is onboarded to prow but still needs the labels managed by gardener-robot, e.g., because it still runs concourse jobs on PRs.
3. List of labels that are needed if the repository only enables the trigger plugin but not the rest of prow, i.e., it only uses prow to run tests.

Before this PR, the set of default labels drifted between repositories. This is now cleaned up.
As a consequence, every repository using the list 1 or 2, gets the `ipcei/*` and `topology/*` labels.

Also, gardener/gardener-extension-registry-cache, gardener/gardener-extension-shoot-rsyslog-relp, gardener/gardener-discovery-server are fully onboarded to prow and use list 1. This will delete all labels managed by gardener-robot after merging this PR.

Fixes #

I checked the diff of the change using these commands:
```
yq 'explode(.) | del(.x-default-labels) | del(.x-trigger-only-labels) | del(.x-labels-keep-gardener-robot)' config/prow/labels.yaml > /tmp/after.yaml
git checkout master
yq 'explode(.)' config/prow/labels.yaml > /tmp/before.yaml

dyff between --omit-header --ignore-order-changes /tmp/{before,after}.yaml
```

### DIFF
--- a/config/prow/labels.yaml
+++ b/config/prow/labels.yaml
@@ -6,5564 +6,1679 @@
 #     target: one of issues, prs, or both (also TBD)
 #     addedBy: human? prow plugin? other?
 ---
+
+# List of labels that are used if the repository is fully onboarded to prow.
+x-default-labels: &default-labels
+# area
+- name: area/audit-logging
+  color: 0052cc
+  description: Audit logging related
+  target: both
+  prowPlugin: label
+  addedBy: anyone
+- name: area/auto-scaling
+  color: 0052cc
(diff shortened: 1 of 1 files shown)


oliver-goetz:
yeah 🥳  thanks a lot 😄 

/lgtm
/approve

PR #2637, opened 2024-11-06
Clean up test logs from `gardener-prow` GCS bucket

oliver-goetz:
/kind enhancement

This PR deletes test logs from `gardener-prow` GCS bucket which are older than 6 months. 
It deletes the logs from `artifacts` directory only. This includes logs created by our kind based e2e tests. Those logs a quite big and deleting them saves storage space and money. 
The prow jobs themselves and their output are kept.

Fixes #

/cc @timebertt


oliver-goetz:
> Nice! Looking forward to the result and seeing how long the job will run 😄

Will it be hours or days? 😉

PR #2644, opened 2024-11-07
Add more shared renovate config presets

timebertt:
/kind bug

This PR adds more shared renovate config presets for config that we see in multiple gardener repositories:

- `makefile-versions`: Regex manager for updating `_VERSION` and `_version` variables in Makefiles and scripts
- `imagevector` (parametrized): Regex manager for detecting dependencies in image vector files matched by the passed regex

The `makefile-versions` preset is used in this repository right away.

Fixes #


oliver-goetz:
Thanks again 😄 

/lgtm
/approve

PR #2717, opened 2024-11-20
Don't protect branches created by dependatbot or renovate

oliver-goetz:
/kind bug

We should not protect branches created by dependabot or renovate even when their branches match include rules.

Fixes #

/cc @acumino


acumino (approved):
Thank You!
/lgtm
/approve

PR #2807, opened 2024-12-03
Add a prometheus instance for collecting performance metrics

timebertt:
/kind enhancement

This PR adds another `Prometheus` object with the remote write receiver enabled in the prow build cluster.

Test jobs can push any performance metrics they want to persist over time to this instance by adding a remote write config to `http://prometheus-performance.monitoring.svc.cluster.local:9090/api/v1/write`.
Alternatively, the `prometheus-performance.prow.gardener.cloud.local` name is rewritten to the service in-cluster name by the prow build cluster's coredns. This is helpful for test jobs that run a kind cluster in the test pod, where a `*.svc.cluster.local` hostname from the outer cluster cannot be resolved within the inner cluster.

Furthermore, this PR allows unauthenticated users to use the `Explore` feature in Grafana, similar to https://github.com/gardener/gardener/pull/4140.

Part of the 2024-12 Gardener Hackathon

https://github.com/gardener/gardener/pull/10964 is a proof-of-concept for pushing metrics to this instance

There is no visualization of the performance metrics yet.

### DIFF
--- a/config/prow/cluster/coredns/coredns-custom-config.yaml
+++ b/config/prow/cluster/coredns/coredns-custom-config.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coredns-custom
+  namespace: kube-system
+data:
+  rewrites.override: |
+    rewrite stop {
+      # Add a local hostname that is rewritten to the in-cluster service name of the prow cluster.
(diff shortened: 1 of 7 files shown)


timebertt:
This is ready, @oliver-goetz helped to verify this change on the prow cluster.

oliver-goetz:
Nice 🚀 

/lgtm
/approve

PR #2812, opened 2024-12-03
Add initial flux deployment

timebertt:
/kind enhancement

**What this PR does / why we need it**:

We want to deploy prow and the additional components in this repository via flux.

This PR adds the initial flux deployment, i.e., only the flux components themselves without any workload application.
Flux manages itself, i.e., the `flux-system` kustomization contains the flux manifests and the flux `Kustomization` for itself as well.

Later on, we can add more `Kustomizations` and `HelmReleases` to deploy additional components and remove them step by step from `deploy.sh`.

**Which issue(s) this PR fixes**:
Part of https://github.com/gardener-community/hackathon/tree/main/2024-12_Schelklingen

**Special notes for your reviewer**:

**Manual steps for rollout**:

After merging this PR, we need to bootstrap the flux self-management by applying the `flux-system` bootstrap kustomization once:

```
# g target --shoot prow-trusted
k apply --server-side -k clusters/prow-trusted
# g target --shoot prow-build
k apply --server-side -k clusters/prow-build
```

PR #3009, opened 2025-01-10
Migrate deployment of `monitoring` stack to `flux` 

oliver-goetz:
/kind enhancement

This is the final step on our journey from deploying the components of the prow clusters via a custom `deploy.sh` to a deployment with `flux`.

This PR migrates the prow monitoring stack to `flux`. Unlike the other changes this also includes a switch from [`kube-prometheus`](https://github.com/prometheus-operator/kube-prometheus) to [`kube-prometheus-stack`](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) helm chart.  
It is easier to deploy `kube-prometheus-stack` via `flux`, updated more frequently and you can configure it in a less resource consuming way. 

Finally, `deploy.sh` and the `deployer` service account have been dropped too, because we don't need them anymore 😄 

Follow-up to https://github.com/gardener/ci-infra/pull/2812

/cc @timebertt 

The new setup is already manually deployed to our prow clusters.


clusters/prow-build/monitoring-build-kustomization.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: monitoring-build
```
timebertt, line 4: In our setup, there is a single `monitoring` `Kustomization` in `clusters/base` and a patch in both `clusters/prow-{build,trusted}` that overwrite `spec.path` to `deploy/monitoring/{build,trusted}`.
This way, the main part of the `Kustomization` is shared/reused and both have the same name for consistency.

This doesn't mean we have to do this here in the same way. Just some ideas :)

deploy/monitoring/build/patch-kube-prometheus-stack.yaml
```diff
@@ -0,0 +1,41 @@
...
+  values:
+    alertmanager:
+      templateFiles:
+        cluster-name.tmpl: |-
+          {{ define "cluster_name" }}Build Cluster{{ end }}
+
+    grafana:
+      ingress:
+        annotations:
+          cert.gardener.cloud/issuer: ci-issuer
```
timebertt, line 16: These annotations are partly duplicated for both clusters. Should we move them to the base?

config/prow/plugins.yaml
```diff
@@ -193,7 +193,7 @@ config_updater:
       clusters:
         gardener-prow-trusted:
         - prow
-    config/prow/cluster/webserver-static-files/*:
+    config/prow/webserver-static-files/*:
```
timebertt, line 196: This is the only deployment-related thing left in the `config/prow` directory.
We could remove the old deployment of the `webserver-static-files` folder via the `config_updater` plugin and deploy it with flux as well :)

timebertt (commented):
Very nice, love it! Thanks for finishing the last item of this effort 🚀 
I don't have much to comment :)

clusters/prow-build/monitoring-build-kustomization.yaml
oliver-goetz: Done 😄

deploy/monitoring/build/patch-kube-prometheus-stack.yaml
oliver-goetz: I first thought about this too. It would save like 3 lines, but the ingress configurations are a bit harder to understand, because you have to look into two different files.
The base file is quite packed, so in my opinion it is easier to understand if you can see the ingress configuration at a glance.

timebertt: I understand your reasoning. Makes sense 👍

timebertt (approved):
Very nice, thanks again!

Leaving a 
/hold
in case you want to watch/control the rollout of this change closely. Just remove it when you're ready :)

PR #3179, opened 2025-02-11
Fix required_status_checks for cert-management

MartinWeindel:
/kind enhancement

Fix required_status_checks for cert-management after switching to multi-arch builds.

Fixes #


marc1404 (approved):
Thank you :) I wasn't aware of this before

/lgtm

PR #3237, opened 2025-02-19
Add fallbacks for gardener tests manipulating /etc/hosts.

ScheererJ:

Add fallbacks for gardener tests manipulating /etc/hosts.

Manipulating /etc/hosts within prow test jobs is prone to races with kubelet creating /etc/hosts in side car containers (see https://github.com/gardener/gardener/issues/11410 for details). Therefore, this change adds several fallbacks for the default domains so that the tests get the correct response even if coredns is queried.

Fixes https://github.com/gardener/gardener/issues/11410

### DIFF
--- a/deploy/coredns/coredns-custom-config.yaml
+++ b/deploy/coredns/coredns-custom-config.yaml
@@ -12,3 +12,28 @@ data:
       # cluster in the test pod because the domain overlaps with the cluster domain of the inner (kind) cluster.
       name exact prometheus-performance.prow.gardener.cloud.local prometheus-performance.monitoring.svc.cluster.local
     }
+  # Manipulating /etc/hosts within prow test jobs is prone to races with kubelet creating /etc/hosts in side car containers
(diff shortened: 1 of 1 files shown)


rfranzke (commented):
/lgtm
/cc @oliver-goetz 

After this PR is merged, we can remove https://github.com/gardener/gardener/blob/c009466cb7e5a92b2df77f24e7fbf85a318283d0/hack/ci-common.sh#L97-L105, right?

ScheererJ:
> /lgtm /cc @oliver-goetz
> 
> After this PR is merged, we can remove https://github.com/gardener/gardener/blob/c009466cb7e5a92b2df77f24e7fbf85a318283d0/hack/ci-common.sh#L97-L105, right?

I would prefer to keep it in place to handle the 99% case and let only the rare race condition be handled through this custom config.

oliver-goetz:
Crazy, not all IPs are as I expected them 😅 

/lgtm
/approve

PR #3250, opened 2025-02-21
Add `runc@v1.24` into `krte` image

oliver-goetz:
/kind enhancement

This PR adds `runc@v1.24` into `krte` image. 
Currently it is download on every test ([ref](https://github.com/gardener/gardener/blob/master/pkg/provider-local/node/patch-runc.sh)), but this is not stable. Thus, if we add the runc binary into `krte` image, we can copy it from disk when testing.

Fixes #

/cc @marc1404 @rfranzke

### DIFF
--- a/images/krte/Dockerfile
+++ b/images/krte/Dockerfile
@@ -92,7 +92,11 @@ RUN echo "Installing Packages ..." \
         && sed -i 's/ulimit -Hn/# ulimit -Hn/g' /etc/init.d/docker \
     && echo "Ensuring Legacy Iptables ..." \
         && update-alternatives --set iptables /usr/sbin/iptables-legacy \
-        && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
+        && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy \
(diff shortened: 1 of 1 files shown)


images/krte/Dockerfile
```diff
@@ -92,7 +92,11 @@ RUN echo "Installing Packages ..." \
         && sed -i 's/ulimit -Hn/# ulimit -Hn/g' /etc/init.d/docker \
     && echo "Ensuring Legacy Iptables ..." \
         && update-alternatives --set iptables /usr/sbin/iptables-legacy \
-        && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
+        && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy \
+    && echo "Getting runc v1.2.4 ..." \
```
marc1404, line 96: Could you also add a `TODO` comment so that we remember to remove it again eventually? :)

marc1404 (commented):
Awesome, thanks! :)
/lgtm

rfranzke:
Awesome!

marc1404 (approved):
Thank you!
/lgtm

PR #3982, opened 2025-06-23
Replicate Perses and Perses Operator images to Gardener GCR

rickardsjp:
/kind enhancement

Adds the Perses and Perses Operator images to GCR. They are hosted on docker hub and need to be replicated.

n/a

cc @ialidzhikov

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -164,6 +164,17 @@ images:
   destination: europe-docker.pkg.dev/gardener-project/releases/3rd/opentelemetry-operator/opentelemetry-operator
   tags:
   - v0.127.0
+- source: persesdev/perses
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/persesdev/perses
+  tags:
+  - v0.50.3
+  - v0.51.0
+- source: persesdev/perses-operator
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/persesdev/perses-operator
+  tags:
+  - v0.1.10
+  - v0.1.12
+  - v0.2.0
 # gardener/pkg/provider-local/images.yaml
 - source: kindest/local-path-provisioner
   destination: europe-docker.pkg.dev/gardener-project/releases/3rd/kindest/local-path-provisioner


config/images/images.yaml
```diff
@@ -164,6 +164,18 @@ images:
   destination: europe-docker.pkg.dev/gardener-project/releases/3rd/opentelemetry-operator/opentelemetry-operator
   tags:
   - v0.127.0
+- source: persesdev/perses
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/persesdev/perses
+  tags:
+  - v0.50.3
```
vpnachev, line 170: I see gardener is using only version v0.51.0, why do you replicate older unused version?

config/images/images.yaml
```diff
@@ -164,6 +164,18 @@ images:
...
+  - v0.51.0
+- source: persesdev/perses-operator
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/persesdev/perses-operator
+  tags:
+  - v0.1.10
+  - v0.1.11
```
vpnachev, line 176: I see gardener is using only version v0.51.0, why do you replicate older unused versions?

config/images/images.yaml
```diff
@@ -164,6 +164,18 @@ images:
...
   - v0.127.0
+- source: persesdev/perses
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/persesdev/perses
+  tags:
+  - v0.50.3
+  - v0.51.0
+- source: persesdev/perses-operator
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/persesdev/perses-operator
+  tags:
+  - v0.1.10
```
ialidzhikov, line 175: Let's only replicate tags which are used in the master branch and in the release-v1.121 branch.

config/images/images.yaml
rickardsjp: Because on an older version of Gardener, we used `v0.50.3`

rickardsjp: https://github.com/gardener/gardener/blob/release-v1.121/imagevector/containers.yaml

config/images/images.yaml
```diff
@@ -164,6 +164,16 @@ images:
...
+  - v0.2.0
```
vpnachev, line 176: gardener/gardener@master is at v0.1.12, it is also not updated to v0.2.0 with https://github.com/gardener/gardener/pull/12371/files#diff-d880f5ed5b194cea49bceb3f252990d7a8476de0027561917fca0452ebd2e150R404, so maybe you want ?

```suggestion
  - v0.1.12
```

rickardsjp: Understood. Then I would prefer to also replicate the v0.1.12 image. I'll add it back so that we have 0.1.10, 0.1.12, and 0.2.0 for the operator.

ialidzhikov (approved):
/lgtm

PR #4153, opened 2025-07-18
registry-cache: Increase the memory requests and limits for the unit tests

dimitar-kostadinov:
/kind enhancement

Increases memory requests and limits for `registry-cache` unit tests jobs to prevent `OOMKilled` failures. 

Fixes https://github.com/gardener/gardener-extension-registry-cache/issues/430

/cc @ialidzhikov


ialidzhikov:
Let's check the usage after https://github.com/gardener/gardener-extension-registry-cache/pull/416 and adapt if needed. We could avoid adapting the resource requests and limits several times. 

/hold
until https://github.com/gardener/gardener-extension-registry-cache/pull/416 is merged.

PR #4279, opened 2025-08-13
rsyslog-relp: Increase the memory requests and limits for the unit tests

dimitar-kostadinov:
/kind enhancement

Increases memory requests and limits for `rsyslog-relp` unit tests jobs to prevent `OOMKilled` failures and `golangci-lint` timeouts.
image

Fixes #

Similar to #4153 
/cc @ialidzhikov @plkokanov


ialidzhikov (commented):
/approve

PR #4402, opened 2025-09-03
Copy `envoyproxy/envoy:distroless-v1.35.0`

oliver-goetz:

/kind task

`distroless-v1.35.1` has a bug and is not non-root. Thus, we copy `distroless-v1.35.0` to be able to change the image source.

Fixes #

/cc @marc1404

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -124,6 +124,7 @@ images:
 - source: envoyproxy/envoy
   destination: europe-docker.pkg.dev/gardener-project/releases/3rd/envoyproxy/envoy
   tags:
+  - distroless-v1.35.0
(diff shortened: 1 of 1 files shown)


config/images/images.yaml
```diff
@@ -124,6 +124,7 @@ images:
 - source: envoyproxy/envoy
   destination: europe-docker.pkg.dev/gardener-project/releases/3rd/envoyproxy/envoy
   tags:
+  - distroless-v1.35.0
   - distroless-v1.35.1
```
marc1404, line 128: I assume `distroless-v1.35.1` will be removed once we adapted the reference in `g/g`?

marc1404 (approved):

PR #4483, opened 2025-09-16
Enhance copy-images to overwrite destination tags

vpnachev:
/kind enhancement

Enhance copy-images to overwrite destination tags

See https://github.com/gardener/gardener/pull/12941

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -6,219 +6,223 @@ images:
 - source: grafana/grafana
   destination: europe-docker.pkg.dev/gardener-project/releases/3rd/grafana/grafana
   tags:
-  - 7.5.17
+  - source: 7.5.17
 # DO NOT ADD NEW COREDNS IMAGES.
 # With https://github.com/gardener/gardener/pull/8192 a GCR copy image is no longer used for coredns.
 - source: coredns/coredns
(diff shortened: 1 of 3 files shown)


config/images/images.yaml
```diff
@@ -6,316 +6,320 @@ images:
...
-  - distroless-v1.35.2
-  - distroless-v1.35.3
+  - source: distroless-v1.35.0
+    destination: v1.35.0-distroless
+  - source: distroless-v1.35.1
+    destination: v1.35.1-distroless
+  - source: distroless-v1.35.2
+    destination: v1.35.2-distroless
+  - source: distroless-v1.35.3
+    destination: v1.35.3-distroless
```
oliver-goetz, line 135: I'm not sure if it we can make the update automation in renovate work with a reasonable effort ([ref](https://github.com/gardener/ci-infra/blob/59c7088bf6121ce39ede5e540248e40804f07732/.github/renovate.json5#L147-L160)) 🤔

oliver-goetz (commented):
Thanks for looking into this 🙂 

The copying itself should work. However, I'm not so sure about the renovate rule which adds new dependencies automatically.

vpnachev:
While I have some ideals for the `source` tag, I cannot come up with solution how renovate to manage the `destination` tag automatically, therefore I will close the PR.

PR #4563, opened 2025-09-30
Skip `gardener` e2e tests if only `OWNERS` or `OWNERS_ALIASES` change

oliver-goetz:

/kind enhancement

Those files might be updated in a higher frequency in the future, so it is worth to consider them in the skip condition as well.

Fixes #


timuthy (commented):
Thank you! 👍 
/lgtm
/approve

PR #4828, opened 2025-11-12
Increase CPU and memory requests for gardenadm e2e tests

oliver-goetz:

/kind enhancement

We would like to test if we can reduce flakes by running each gardenadm e2e test alone on a node.

Fixes #

/cc @ScheererJ


ScheererJ (approved):
/lgtm
/approve

PR #4926, opened 2025-11-27
Use go build cache for tests

shafeeqes:

/kind enhancement

This is a test implementation of github.com/saracen/gobuildcache for our e2e, integration and unit tests.
We need to observe the test runs and gain some confidence, hence the `time` command.

The pre-submit jobs uses a read-only cache and the periodic tests use the read-write mode.

The test results themselves are not cached, see [docs](https://pkg.go.dev/cmd/go?ref=500.keboola.com#hdr-Testing_flags:~:text=If%20a%20run%20of%20go%20test%20has%20any%20test%20or%20non%2Dtest%20flags%20outside%20this%20set%2C%20the%20result%20is%20not%20cached) and we use these two flags for junit reporting and timeout https://github.com/gardener/gardener/blob/28a432104beec0235fa7e35c492ce2a4d4298b22/hack/test-integration.sh#L21-L23. But we could still benefit from the build cache.

Co-Authored-By @Shegox  at the [2025-11 Hackathon](https://gardener.cloud/community/hackathons/2025-11/).

Sharing some initial results for the `e2e-kind` test:
The first run:
```
real	84m44.456s
user	61m13.853s
sys	6m51.593s
```

With cache:
```
real	81m42.910s
user	2m49.160s
sys	1m10.514s
```
```
real	80m31.979s
user	2m54.775s
sys	0m59.914s
```
```
real	83m48.854s
user	3m25.234s
sys	1m8.477s
```
Running the test again without enabling cache:
```
real	86m24.553s
user	66m28.328s
sys	7m3.101s
```

We can see that the clock time is still same, but the CPU time in the user space is significantly lesser because of the build cache.

Fixes #

This is a rework of https://github.com/gardener/ci-infra/pull/4903.
Prow overwrites the GOPATH, hence the additional PATH setting, this can be improved later once we build `gobuildcache` as part of the test image in gardener.

Workload identity federation is used to provide access to the GCS bucket. Documentation will follow.

### DIFF
--- a/config/jobs/gardener/gardener-e2e-kind-ha-multi-node-upgrade.yaml
+++ b/config/jobs/gardener/gardener-e2e-kind-ha-multi-node-upgrade.yaml
@@ -13,17 +13,24 @@ presubmits:
     labels:
       preset-dind-enabled: "true"
       preset-kind-volume-mounts: "true"
+      preset-gocache-mounts: "true"
     annotations:
       description: Runs end-to-end tests on a Seed with single zone with a Shoot with failure tolerance type 'node' for gardener developments in pull requests
       fork-per-release: "true"
     spec:
+      serviceAccountName: gardener-prow-gobuildcache-ro
       containers:
       - image: europe-docker.pkg.dev/gardener-project/releases/ci-infra/krte:v20251126-51275a0-1.25
         command:
         - wrapper.sh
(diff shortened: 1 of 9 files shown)


oliver-goetz (approved):
Here we go again 😄

PR #4991, opened 2025-12-08
Move back to public.ecr.aws registry

DockToFuture:

/kind enhancement

Move back to `public.ecr.aws` registry as it is now reachable via AAAA record.
Fixes #

See: https://github.com/gardener/gardener-extension-provider-aws/pull/1597


oliver-goetz (approved):
Thanks for cleaning up 🧹

PR #5004, opened 2025-12-10
config/images: add opentelemetry target allocator and collector images

dnaeon:

/kind enhancement

This PR vendors the OpenTelemetry Target Allocator image. Also it adds `0.140.0` and `0.141.0`  versions of the `opentelemetry-collector-contrib` image.

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -217,6 +217,13 @@ images:
   - 0.127.0
   - 0.129.1
   - 0.139.0
+  - 0.140.0
+  - 0.141.0
+- source: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/opentelemetry-operator/target-allocator
+  tags:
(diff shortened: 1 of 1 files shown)


config/images/images.yaml
```diff
@@ -216,6 +216,16 @@ images:
...
+- source: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/opentelemetry-operator/target-allocator
+  tags:
+  - v0.140.0
+  - v0.141.0
+- source: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/opentelemetry-operator/opentelemetry-collector
+  tags:
+  - 0.140.0
+  - 0.141.0
```
rfranzke: Do we need this or can we just fall back to always using the `contrib` images (which also seems to have support for extensions that we might need (bearer token auth))?

PR #5009, opened 2025-12-11
Remove etcd-druid unit & integration tests from running in prow

anveshreddy18:

/kind cleanup
/kind task

This PR removes the prow jobs `etcd-druid-unit`, `etcd-druid-api-unit` and `etcd-druid-integration` from the `etcd-druid` jobs configuration as we've moved these to run with GHA in [this PR](https://github.com/gardener/etcd-druid/pull/1229) and we do not want to have multiple jobs running the same tests across GHA & prow. Hence this removal.

**Special notes for your reviewer**: NA


shreyas-s-rao (approved):
@anveshreddy18 thanks for raising the PR. Confirmed that the GHA test steps work fine on PRs, so we don't really need these 3 presubmit jobs anymore.
/lgtm

oliver-goetz (commented):
May I ask what's the reason for this change?

We have multiple tasks ongoing to enroll prow to the entire gardener organization to have a unified developer experience. 
This change goes the opposite direction.

anveshreddy18:
@oliver-goetz, In addition to what @shreyas-s-rao mentioned, we are also doing the same with `etcd-backup-restore` wherein we have the checks, unit & integration tests running through GHA and we are running kind based tests in prow. We thought that's the way to go as very recently we were onboarded onto GHA across all the etcd repos. 

But if you say that's not the case and would need these non resource intensive tests also to be running in prow for other reasons, I'm good with it, we'll close this. We just didn't have the full picture on the way forward and would like you to educate us on that. Thanks.

oliver-goetz:
My understanding about the midterm scenario is:
- we use prow for testing, github interactions (`/lgtm`, `/approve`, ...) and merging PRs
- we use Github Actions to build artifacts (docker images, helm charts, ocm descriptors)

It will take some time until all repos get there, but we are preparing things for it. Like Dimitar who created lots of PRs to add OWNERS files to the repos like [this one for etcd-druid](https://github.com/gardener/etcd-druid/pull/1184).

I prefer to keep all the tests in prow.
Otherwise, you'll observe different testing behavior when we activate tide for merging PRs. In this case prow tests a PR with the latest state of main branch before merging, while Github Actions do not run such tests.

oliver-goetz:
> > we use Github Actions to build artifacts
> 
> We would like to run tests against a commit before publishing images for it, because we don't want to publish possibly buggy images. Hence the dependency for image builds on tests, hence we need to run the tests in GHA. If at all GHA image build action can depend on a prow job result like the unit test, then that would be great for us to get rid of the tests running in GHA. If not, we'll need to mandatorily run the tests in GHA.

In `gardener/gardener` we don't build artifacts for PRs, since nobody is using them anyway.
A [check_run](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#check_run) and a status could be potential triggers for GHA actions too. With this it might be possible to trigger a build GHA when another test has been successful. I never tried.

We only build when when an PR is merged to the main or a release branch. In this case, prow tested it before (rebased to the latest state of main branch). Thus, testing it again before building the images would bring no additional value.

> 
> > I prefer to keep all the tests in prow.
> 
> We can also keep it in prow for now, but due to the above constraint I mentioned, we cannot remove those from GHA. So we'll end up having few duplicate tests running for each PR.

So you do not push a broken image, but you have a broken main branch.
When tide is used for merging PRs, it will run all tests before merging again, so you won't have a broken main branch in the first place.
I understand your issue. In my opinion it is a work around that GHA do not test again before merging.

For images built based on PRs there is no guarantee anyway. I could also disable all tests before pushing.

> 
> > Otherwise, you'll observe different testing behavior when we activate tide for merging PRs. In this case prow tests a PR with the latest state of main branch before merging, while Github Actions do not run such tests.
> 
> Does tide check whether all the PR checks have passed, or only the ones executed by prow?

tide respects other test results too.

PR #5023, opened 2025-12-12
Split autobump for prow components and jobs in a better way

oliver-goetz:

/kind enhancement

Today, almost all prow components and prow jobs are auto-bumped by the same job.
Because of this, it is hard to find out when actual prow components have been updated (e.g. to check if there is a regression). 
There is already an own autobump config for prow jobs which is rarely used at the moment.

With this PR autobumps for prow components and for prow jobs (and related stuff) are separated more clearly between both autobump configs. 

Fixes #

/cc @marc1404 @LucaBernstein


marc1404 (approved):
Great idea! :)

PR #5074, opened 2025-12-18
Fix grafana datasources not being loaded properly

oliver-goetz:

/kind bug

Datasources in `grafana` are not loaded properly which seems to be a bug in the `k8s-sidecar`. 
See also https://github.com/grafana/helm-charts/issues/4023#issuecomment-3633527427

Fixes #

/cc @shafeeqes


shafeeqes:
Thank you!
/lgtm
/approve

PR #5127, opened 2026-01-07
Reduce CPU and memory requests for gobuildcache enabled tests

shafeeqes:

/kind enhancement

After https://github.com/gardener/ci-infra/pull/4926, we have seen a consistent reduction in the CPU and memory usage for the e2e test. Let's reduce the requests so that the nodes can be better utilized.

For eg, for `ci-gardener-e2e-kind`,
without cache,

image
image

With cache,
image
image

Fixes #

### DIFF
--- a/config/jobs/gardener/gardener-e2e-kind-ha-multi-node.yaml
+++ b/config/jobs/gardener/gardener-e2e-kind-ha-multi-node.yaml
@@ -37,8 +37,8 @@ presubmits:
           privileged: true
         resources:
           requests:
-            cpu: 12
-            memory: 48Gi
+            cpu: 6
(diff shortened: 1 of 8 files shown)


dimityrmirchev (commented):
Thanks!

Should these be also reduced for other e2e jobs, e.g. https://github.com/gardener/ci-infra/blob/819819ac71303fec36b3bfde7d402a9d767ba1f3/config/jobs/gardener-discovery-server/gardener-discovery-server-e2e-kind.yaml#L26-L29? Does it make sense to track such initiative centrally?

shafeeqes:
/hold
To clear some doubts

PR #5317, opened 2026-02-04
Add new crypto label

dimityrmirchev:

/kind enhancement
/area security

Adds a new `crypto` label which is going to be used to ensure visibility and review of crypto-impacting changes.

/cc @JordanJordanov


oliver-goetz:
Thanks 👍 

/lgtm
/approve

PR #5924, opened 2026-04-10
Fix tls certificates for `oauth2-proxy`

oliver-goetz:

/kind bug

Before the domains for the certificate request were automatically collected from `Gateway` and `HTTPRoute`. This does not work because you cannot include a wildcard domain (`*.prow.gardener.cloud`) and a matching subdomain (`oauth2.prow.gardener.cloud`). in the same letsencrypt request.

Fixes #

/cc @MartinWeindel 
Thanks 🙂


MartinWeindel (commented):
lgtm

marc1404 (approved):
Thanks! :)

PR #6081, opened 2026-04-28
Fix falco service unit test

AleksandarSavchev:

/kind bug

The `make verify-extended` command in the falco extension also uses some `python` code. This caused the unit test to fail with python errors: https://prow.gardener.cloud/view/gs/gardener-prow/pr-logs/pull/gardener_gardener-extension-shoot-falco-service/498/pull-gardener-extension-shoot-falco-service-unit/2049087496330940416#

Changed the unit test to use the `.ci/verify` script, that installs required `python` packages: https://github.com/gardener/gardener-extension-shoot-falco-service/blob/main/.ci/verify

Fixes #


oliver-goetz (approved):
Thanks 🙂

PR #6215, opened 2026-05-19
Pin docker-ce to `29.5.0` in krte image to avoid Docker `29.5.1` regression

shafeeqes:

/kind bug

Docker 29.5.1 introduced a regression (moby/moby#52653) where `docker cp` fails with "mkdirat var/run: file exists" (see [failed job](https://prow.gardener.cloud/view/gs/gardener-prow/pr-logs/pull/gardener_gardener/14863/pull-gardener-e2e-kind-release-v1-142/2056674180337766400#1:build-log.txt%3A331)) on containers that have `/var/run` as a symlink to `/run` (as kind node containers do). This breaks all non-upgrade kind e2e tests at the `make kind-single-node-up` step.

Pin docker-ce to `5:29.5.0-1~debian.12~bookworm` until Docker `29.5.2` ships with the fix. 

Fixes #

### DIFF
--- a/images/krte/Dockerfile
+++ b/images/krte/Dockerfile
@@ -88,7 +88,10 @@ RUN echo "Installing Packages ..." \
             "$(. /etc/os-release && echo "$VERSION_CODENAME")" stable" | \
             tee /etc/apt/sources.list.d/docker.list > /dev/null \
         && apt-get update \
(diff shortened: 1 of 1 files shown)


ary1992 (commented):
/lgtm
/approve

PR #6216, opened 2026-05-19
Pin `krte` image to the last version which is using Docker v29.5.0

oliver-goetz:

/kind bug

Docker [v29.5.1 ](https://docs.docker.com/engine/release-notes/29/#2951)introduced a regression which makes all our e2e tests fail.

Thus, this PR pins the  `krte` image to the latest version which is using Docker v29.5.0 until v29.5.2 was released.

Fixes #

/cc @ScheererJ


oliver-goetz:
There is #6215 which fixes the issue too 🙂 

/close

PR #6252, opened 2026-05-22
chore: remove obsolete gardener-extension-shoot-networking-filter-e2e…

domdom82:
…-operator test

/kind cleanup

- Since g/g 1.142.0 gardener-operator is the default, there is no need for a separate test
- Existing e2e test has been made to use gardener-operator, removing the extra operator-test here


oliver-goetz (approved):
🧹 

/lgtm
/approve

PR #6578, opened 2026-07-02
Add static peribolos config checker

StarMiner99:

/kind enhancement

This PR adds a static configuration checker for peribolos (for the org repo).
The tool is able to check the config/org.yaml for various mistakes, see README.md for elaboration.

Related [org/#36](https://github.com/gardener/org/issues/36)
Will create a follow up with prow job -> will be opened as soon as this is merged and a docker image with the peribolos-checkconfig tool is published to the registry.

Will the docker image push succeed? Is further configuration required in the docker registry?

### DIFF
--- a/prow/cmd/peribolos-checkconfig/README.md
+++ b/prow/cmd/peribolos-checkconfig/README.md
@@ -1,0 +1,73 @@
+# peribolos-checkconfig
+
+`peribolos-checkconfig` is a static linter for peribolos org
+configuration files. It validates a peribolos `org.yaml` against a set of
+consistency rules **without needing a GitHub token** — making it a natural fit
+for a presubmit prow job on the repo that stores your org config.
+
+## What it checks
+
+Given a peribolos `org.FullConfig`, for each org it verifies:
+
+- **YAML is well-formed** and unmarshals cleanly into the peribolos schema.
+- **Minimum admins** — each org has at least `--min-admins` admins (default `5`,
+  minimum `2`).
+- **Required admins present** — every user passed via `--required-admins`
+  appears in the org's `admins` list.
+- **No dual-role users** — no user is listed as both an `admin` and a `member`
+  of the same org. Comparison is done after `github.NormLogin` so casing
+  differences (e.g. `Alice` vs `alice`) are caught.
+- **Team members are org members** — every user listed as a team `member` or
+  `maintainer` (recursively across `children`) is also listed at the org level
+  as `admin` or `member`.
+- **Unique team names** — team `name` + all entries in `previously` are unique
+  within an org.
+- **Nested teams are `closed`** — any team that has a parent or has children
+  must have `privacy: closed`. Teams with unset `privacy` are allowed (peribolos
+  fills in `closed` at apply time).
+- **Unique repo names** — repo `name` + all entries in `previously` are unique
+  within an org, compared case-insensitively (GitHub repo names are
+  case-insensitive).
+- **No `archived: false`** — repos configured with `archived: false` are
+  rejected, because the GitHub API cannot unarchive a repository.
+
+Checks that require live GitHub state (`--maximum-removal-delta`, `require-self`,
+current-org reconciliation, invitation flow) are intentionally out of scope —
+they belong to peribolos itself.
+
+## Usage
+
+```
+peribolos-checkconfig \
+  --config-path=./org.yaml \
+  --min-admins=5 \
+  --required-admins=alice --required-admins=bob
+```
+
+### Flags
+
+| Flag                 | Default | Description                                                |
+| -------------------- | ------- | ---------------------------------------------------------- |
+| `--config-path`      | *(required)* | Path to peribolos org config YAML.                    |
+| `--min-admins`       | `5`     | Fail if any org has fewer admins than this. Must be ≥ 2.   |
+| `--required-admins`  | *(none)* | Repeatable; usernames that must appear as admins in every org. |
+| `--log-level`        | `info`  | One of `panic`, `fatal`, `error`, `warn`, `info`, `debug`, `trace`. |
+
+### Exit codes
+
+- **0** — all orgs pass every check.
+- **non-zero** — one or more checks failed; the failing checks are aggregated
+  and logged before exit.
+
+
+## Development
+
+```bash
+# Build
+go build ./prow/cmd/peribolos-checkconfig
+
+# Run tests
+go test ./prow/cmd/peribolos-checkconfig
+```
+
+See [main_test.go](main_test.go) for the check-by-check test matrix.
--- a/prow/cmd/peribolos-checkconfig/README.md
+++ b/prow/cmd/peribolos-checkconfig/README.md
@@ -17,6 +17,10 @@ Given a peribolos `org.FullConfig`, for each org it verifies:
 - **No dual-role users** — no user is listed as both an `admin` and a `member`
   of the same org. Comparison is done after `github.NormLogin` so casing
   differences (e.g. `Alice` vs `alice`) are caught.
+- **No dual-role team users** — no user is listed as both `member` and
+  `maintainer` of the same team. Peribolos itself only enforces this at apply
+  time when `--fix-team-members` is set, so bad configs can otherwise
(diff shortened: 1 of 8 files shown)


prow/cmd/peribolos-checkconfig/main.go
```diff
@@ -0,0 +1,243 @@
...
+	var o options
+	if err := o.parseArgs(flag.CommandLine, os.Args[1:]); err != nil {
+		logrus.Fatalf("Invalid flags: %v", err)
+	}
+	return o
+}
+
+func (o *options) parseArgs(flags *flag.FlagSet, args []string) error {
+	o.requiredAdmins = flagutil.NewStrings()
+	flags.Var(&o.requiredAdmins, "required-admins", "Ensure config specifies these users as admins")
```
tobschli, line 48: What is the use-case behind ensuring this?

prow/cmd/peribolos-checkconfig/main.go
```diff
@@ -0,0 +1,243 @@
...
+	flags.StringVar(&o.config, "config-path", "", "Path to org config.yaml")
+	flags.IntVar(&o.minAdmins, "min-admins", defaultMinAdmins, "Ensure config specifies at least this many admins")
```
tobschli, line 50: What is the use-case behind ensuring this?

prow/cmd/peribolos-checkconfig/main.go
```diff
@@ -0,0 +1,243 @@
...
+	if n := len(wantAdmins); n < opt.minAdmins {
+		errs = append(errs, fmt.Errorf("%s must specify at least %d admins, only found %d", orgName, opt.minAdmins, n))
+	}
+
+	// check required admins
+	var missing []string
+	for _, r := range opt.requiredAdmins.Strings() {
+		if !wantAdmins.Has(r) {
+			missing = append(missing, r)
+		}
```
tobschli, line 134: nit: whilst member check seems to be normalized below, here, `wantAdmins` is case-sensitive

prow/cmd/peribolos-checkconfig/main.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	orgUsers := normAdmins.Union(wantMembers)
+
+	// team-level checks: collect team members recursively, verify uniqueness of team names
+	teamMembers := sets.Set[string]{}
+	teamNames := sets.Set[string]{}
+	duplicateTeamNames := sets.Set[string]{}
+	nestedNotClosed := []string{}
+	var walkTeams func(teams map[string]org.Team, parentIsNested bool)
+	walkTeams = func(teams map[string]org.Team, parentIsNested bool) {
```
tobschli: Why do you define this recursive function within this `checkOrg` function?

prow/cmd/peribolos-checkconfig/README.md
tobschli, line 1: Per [this](https://github.com/gardener/org/issues/36#issuecomment-4853399800) comment on the issue, we could also semantically validate
> `teams`. One single user could be either enlisted in maintainers or members

I did not see this here. Did I miss it?

prow/cmd/peribolos-checkconfig/main_test.go
tobschli, line 1: Usually, we write tests using the ginkgo testing framework, not with the go testing framework alone.
Would you be able to utilize it?

tobschli (changes requested):
Thank you very much for the PR! ❤️
Just some smaller questions. Otherwise this looks good to me

prow/cmd/peribolos-checkconfig/main.go
StarMiner99: there is a flag within peribolos which fails the peribolos run if specific users are not in the yaml.
see https://docs.prow.k8s.io/docs/components/cli-tools/peribolos/ in the Settings section.

StarMiner99: could prevent peribolos from working so checking this would make sense

prow/cmd/peribolos-checkconfig/main.go
StarMiner99: same as above

prow/cmd/peribolos-checkconfig/main.go
StarMiner99: > In order to mitigate the chance of applying erroneous configs, the peribolos binary includes a few safety checks:
> 
> --required-admins= - a list of people who must be configured as admins in order to accept the config (defaults to empty list)
> --min-admins=5 - the config must specify at least this many admins

prow/cmd/peribolos-checkconfig/main.go
StarMiner99: used the same pattern as here: https://github.com/kubernetes-sigs/prow/blob/ef12e757d9a9ede25c2a41236ace781e56b23265/cmd/peribolos/main.go#L408

peribolos is case-sensitive in this case so checking case insensitive could lead to issues

StarMiner99: (the second check is case-sensitive again (in peribolos))

prow/cmd/peribolos-checkconfig/main.go
StarMiner99: https://github.com/kubernetes-sigs/prow/blob/ef12e757d9a9ede25c2a41236ace781e56b23265/cmd/peribolos/main.go#L765

got it from here.
should i do it iterative instead of recursive?

StarMiner99: the reason for this is that github teams can be nested

prow/cmd/peribolos-checkconfig/README.md
StarMiner99: nope missed it will fix

StarMiner99: added it

prow/cmd/peribolos-checkconfig/main_test.go
StarMiner99: done

prow/cmd/peribolos-checkconfig/main.go
tobschli: It is all good that it is a recursive function, but I wondered if there is a reason on why you define the function inside of the `checkOrg` function, instead of treating it as a "helper" and define it like any other normal function.

StarMiner99: extracted the function with state struct as helper

tobschli (commented):
/lgtm

tobschli:
@oliver-goetz do you also want to have a look over this?

prow/cmd/peribolos-checkconfig/main_test.go
```diff
@@ -0,0 +1,464 @@
...
+	secret := org.Secret
+
+	DescribeTable("#checkOrg",
+		func(opt options, orgName string, orgConfig org.Config, expectError bool) {
+			err := checkOrg(&opt, orgName, orgConfig)
+			if expectError {
+				Expect(err).To(HaveOccurred(), "expected error, got none")
+			} else {
+				Expect(err).ToNot(HaveOccurred(), "unexpected error")
+			}
```
oliver-goetz: You could check for error strings to ensure that the tests do not fail for the wrong reason.

prow/cmd/peribolos-checkconfig/main_test.go
```diff
@@ -0,0 +1,464 @@
...
+		description := "cool repo"
+
+		DescribeTable("validates repo configs",
+			func(config map[string]org.Repo, expectError bool) {
+				err := validateRepos("myorg", config)
+				if expectError {
+					Expect(err).To(HaveOccurred(), "expected error, got none")
+				} else {
+					Expect(err).ToNot(HaveOccurred(), "unexpected error")
+				}
```
oliver-goetz: Same here, please check the error strings.

oliver-goetz (changes requested):
Nice PR thank you 😄 

I found only two little things in the tests.

oliver-goetz (approved):
Thanks 👍

PR #6658, opened 2026-07-14
Increase memory for `external-dns-management-unit-tests.yaml`

MartinWeindel:

/kind enhancement

Increase memory for `external-dns-management-unit-tests.yaml` as it often fails with OOM.

Fixes #

/cc @marc1404


marc1404 (commented):
More is more 🧘‍♂️
/lgtm
/approve

PR #6688, opened 2026-07-16
[renovate] Enable `automerge` for image copies

oliver-goetz:

/kind enhancement

It is a repetitive (often forgotten) task to approve image copies which can be easily automated. 

Fixes #

/cc @marc1404

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -169,6 +169,7 @@
       matchFileNames: [
         'config/images/images.yaml',
       ],
+      automerge: true,
       postUpgradeTasks: {
         commands: [
           'go install github.com/mikefarah/yq/v4@latest',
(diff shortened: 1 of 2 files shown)


marc1404 (commented):
Great idea, I love it! 🫶 
Should we adapt/remove this hint message?
https://github.com/gardener/ci-infra/blob/e3692c920bc6013070c62ec95fdfab1880cf1961/config/renovate/imagevector.json5#L41

I think we shouldn't remove it because depending on the timing the `g/g` PR would be open before the image has been replicated and tests would fail (so a test rerun necessary)

oliver-goetz:
> Great idea, I love it! 🫶 Should we adapt/remove this hint message?
> 
> https://github.com/gardener/ci-infra/blob/e3692c920bc6013070c62ec95fdfab1880cf1961/config/renovate/imagevector.json5#L41
> 
> I think we shouldn't remove it because depending on the timing the `g/g` PR would be open before the image has been replicated and tests would fail (so a test rerun necessary)

Good idea 😄 
I just updated the hints, PTAL.

marc1404 (approved):
Thanks a lot!

PR #6685, opened 2026-07-16
Add owners-aliases-bumper tool to sync OWNER_ALIASES from peribolos config

StarMiner99:
**How to categorize this PR?**

/kind enhancement

**What this PR does / why we need it**:
Adds a prow cmd tool which syncs OWNER_ALIASES by opening PRs:
- only repositories mentioned in the peribolos config are touched.
- the tool will look at the peribolos config in order to decide who will be in the team alias
- every alias that is already present in OWNER_ALIASES will be updated with the members of the Team with the same name. (if no Team exists with this name nothing happens)
- sub teams aliases are as follows: `<parent-team>/<child-team>`

**Which issue(s) this PR fixes**:
Related gardener/org#5 -> needs a post-submit job which runs this tool in order to close this issue. should run on:
- merge to main in org repo if org.yaml is touched) 
- merge to main in any repo (repos which are mentioned in org.yaml) if OWNER_ALIASES is changed
-> will get follow up PR

Related: gardener/org#32 -> influences decisions on team names

**Special notes for your reviewer**:
If https://github.com/kubernetes-sigs/prow/pull/801 gets merged, commit 30b2307ffa2532f3c017ea0e88e588d150cd0d68 can be reverted and simplified -> nothing happened yet, so if it gets merged -> follow up
This tool fits very nicely into post-submits therefore, i decided to make a cmd tool and not a plugin.

PR #6785, opened 2026-07-27
Add prow jobs for owners-aliases-bumper

### LINKED ISSUE #5: Automate maintenance of `OWNERS_ALIASES` files
An automation (probably Prow plugin) should be implemented to generate the content of `OWNERS_ALIASES` files ([example](https://github.com/gardener/gardener/blob/master/OWNERS_ALIASES)).

Specifically, when an `OWNERS_ALIASES` file contains team names that correspond to existing GitHub teams, their members should be automatically added to or removed from the `OWNERS_ALIASES`.

**Why is this needed**:
After adding a declarative org management (see #2), the primary membership source are GitHub teams. Unfortunately, `OWNERS_ALIASES` can't refer to GitHub teams directly which requires manual syncing at the moment.

StarMiner99:

/kind enhancement

Follows #6685
Adds prow jobs for the owners-aliases-bumper tool.

postsubmit -> if org.yaml gets changed we instantly update the aliases
periodic -> if new alias gets added to one of the OWNERS_ALIASES in any repo, tool needs to be run

Fixes gardener/org#5

This tool will not do anything unless the repos get listed in gardener/org config/org.yaml -> Follow up PR in org

### DIFF
--- a/config/jobs/org/org-periodics.yaml
+++ b/config/jobs/org/org-periodics.yaml
@@ -6,7 +6,7 @@ periodics:
   extra_refs:
   - org: gardener
     repo: org
-    base_ref: master
+    base_ref: main
   annotations:
     description: Bumps OWNERS_ALIASES in all configured repos to match the Peribolos config. Periodic run.
     testgrid-create-test-group: "false"
--- a/config/jobs/org/org-periodics.yaml
+++ b/config/jobs/org/org-periodics.yaml
@@ -1,0 +1,34 @@
+periodics:
+- cron: "13 * * * *"
+  name: ci-org-owners-aliases-bumper
+  cluster: gardener-prow-trusted
+  decorate: true
+  extra_refs:
+  - org: gardener
+    repo: org
+    base_ref: master
+  annotations:
+    description: Bumps OWNERS_ALIASES in all configured repos to match the Peribolos config. Periodic run.
+    testgrid-create-test-group: "false"
+  reporter_config:
+    slack:
+      channel: prow-alerts
+  spec:
+    containers:
+    - image: europe-docker.pkg.dev/gardener-project/releases/ci-infra/owners-aliases-bumper:v20260727-9acf867
+      command:
+      - /owners-aliases-bumper
+      args:
+      - --peribolos-conf=config/org.yaml
+      - --github-token-path=/etc/github-token/token
+      - --github-endpoint=http://ghproxy.prow.svc.cluster.local
+      - --github-endpoint=https://api.github.com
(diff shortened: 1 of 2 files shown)


config/jobs/org/org-periodics.yaml
```diff
@@ -0,0 +1,34 @@
+periodics:
+- cron: "13 * * * *"
+  name: ci-org-owners-aliases-bumper
+  cluster: gardener-prow-trusted
+  decorate: true
+  extra_refs:
+  - org: gardener
+    repo: org
+    base_ref: master
```
LucaBernstein:
```suggestion
    base_ref: main
```

config/jobs/org/org-postsubmits.yaml
```diff
@@ -0,0 +1,30 @@
+postsubmits:
+  gardener/org:
+  - name: post-org-owners-aliases-bumper
+    cluster: gardener-prow-trusted
+    branches:
+    - ^master$
```
LucaBernstein:
```suggestion
    - ^main$
```

LucaBernstein (commented):
Thanks for driving this forward 🚀 
`gardener/org` [branch name](https://github.com/gardener/org/branches) needs correction. Else lgtm.

config/jobs/org/org-periodics.yaml
```diff
@@ -0,0 +1,34 @@
+periodics:
+- cron: "13 * * * *"
```
LucaBernstein, line 2: @oliver-goetz wdyt about the frequency? :)

oliver-goetz: My first thought was "this is quite frequent", but now I think it is ok for now. 
The changes in `org` repo are detected by the post submit job, but there is no detection for changes of the `OWNERS_ALIASES` files in the repos. If we run it less frequently, people have to wait longer for an update when they update a group that it fits to a Github team.

oliver-goetz: Ideally, we would have a generator for prow jobs (presubmit and postsubmit), we would like to run in each repo. In this case we could replace the periodic job with a postsubmit template which watches `OWNERS_ALIASES` files in all repos.

StarMiner99: we would need a postsubmit for every repo for this right? a wildcard job would be great 😆: gardener/*

StarMiner99:
~wait for #7053~ merged

LucaBernstein (approved):
Thank you!

PR #6813, opened 2026-07-29
Add Envoy Gateway Image

Michael5601:

/kind enhancement

This PR adds a envoy gateway image and a helm chart to the images.yaml file.

### DIFF
--- a/config/images/images.yaml
+++ b/config/images/images.yaml
@@ -160,6 +160,10 @@ images:
   - distroless-v1.38.2
   - distroless-v1.38.3
   - distroless-v1.39.0
+- source: envoyproxy/gateway
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/envoyproxy/gateway
+  tags:
+  - v1.8.3
 - source: ghcr.io/credativ/vali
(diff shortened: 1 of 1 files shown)


config/images/images.yaml
```diff
@@ -160,6 +160,14 @@ images:
...
   - distroless-v1.38.3
   - distroless-v1.39.0
+- source: envoyproxy/gateway
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/envoyproxy/gateway
+  tags:
+  - v1.8.3
+- source: envoyproxy/gateway-helm
+  destination: europe-docker.pkg.dev/gardener-project/releases/3rd/envoyproxy/gateway-helm
+  tags:
+  - v1.8.3
```
oliver-goetz: We don't copy helm charts here since it is not tested and usually does not add additional value.
Thus, if there is no special reason for copying it please remove it.

oliver-goetz (changes requested):
Thanks for your PR 🙂

config/images/images.yaml
Michael5601: done.

oliver-goetz (approved):
Nice 🙂

PR #6860, opened 2026-08-05
Increase cert-management unit test job memory to 8Gi

marc1404:

/kind enhancement

gardener/gardener v1.148.1 pulled in a large set of new transitive dependencies (VictoriaMetrics stack, Hashicorp memberlist, etc.) that significantly increase compilation memory during `go test -race`.

The pull-cert-management-unit job was OOM-killed at 4Gi. Raise the limit to 8Gi and the request to 6Gi, matching the external-dns-management unit test job which runs the same toolchain and passes at this size.

Fixes: https://github.com/gardener/cert-management/pull/783

/cc @MartinWeindel

### DIFF
--- a/config/jobs/cert-management/cert-management-unit-tests.yaml
+++ b/config/jobs/cert-management/cert-management-unit-tests.yaml
@@ -19,10 +19,10 @@ presubmits:
         - test
         resources:
           limits:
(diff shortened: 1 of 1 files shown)


marc1404 (commented):
/lgtm
/approve

